PR #51, opened 2017-06-16
Added functionality for connecting to kerberos encrypted hive databases

jsoren:


pyongjoo:
When you parse the principal (at VerdictConnection.java:61), I think it would be better to use the existing key, value parsing at the line VerdictConnection.java:53 (after some modification of the rule for value). Can you see if this is possible?

jsoren:
I'm not sure I understand. From what I see, the key-value parsing looks for
strings that look like: "key=value".

However, the principal is of the form "service/host@REALM". I don't see how
I can re-use the key-value parsing here.

pyongjoo:
it's of the form "principal=hive/host@realm", so its key is "principal" and
the value is the right hand side.

pyongjoo:
Also, I think was wrong when I said we should set the "kerberos"
configuration to "true" in VerdictConnection. If the "principal" property
is set by the Properties parameter (instead of being specified in the url
string), dbms instance won't know this.

Can you (1) make the key-value parsing in VerdictConnection able to read
kerberos-related values (host, realm, and so on), and (2) make the dbms
class make a Kerberos connection if there's a "principal" key in
VerdictConf? I guess the realm parsing part can be placed inside the dbms
class (e.g. in the composeUrl() function)?

jsoren:
Okay, I think I implemented this properly.

Now, VerdictConnection will try and find key=value pairs after the ? in the
url (at this point only the principal, but it will add any such pairs it
finds), and adds them to the configuration.

Then, in composeURL, the function checks for the existence of a principal,
and parses the service, host, and realm at that point.

PR #52, opened 2017-06-21
Added support for impala kerberos connection

jsoren:


pyongjoo:
Merging the update for Impala Kerberos connection.

PR #60, opened 2017-07-17
read from Json file

xuxiaoran:
also tested


core/src/main/java/edu/umich/verdict/VerdictConf.java
```diff
@@ -79,6 +91,43 @@ private VerdictConf updateFromStream(InputStream stream) throws FileNotFoundExce
         scanner.close();
         return this;
     }
+    
+    private VerdictConf readFromJson(InputStream stream) throws FileNotFoundException {
```
pyongjoo: (minor) Let's change the method signature to `void updateFromJson(String filepath)`. "read" doesn't sound like it causes any updates. And let this method include the "file-to-stream" part.

PR #69, opened 2017-08-03
redshift connected, able to create meta data, has error on quote string and schema name, will fix in next pull

JunhaoWang:


pyongjoo:
No need to merge work in progress.

PR #93, opened 2017-11-07
Hao

JunhaoWang:
Add lower to ANTLR file


pyongjoo:
This commit includes the .metadata folder, which must not be included in our repository. All those files must be removed from github files.

PR #124, opened 2018-04-04
Upload DbmsMysql.java  ;  Modify Dbms.java

henryqin1997:
Upload DbmsMysql.java  ;  Modify "if else branch" in Dbms.java  ;   maven dependency not solved yet


dongyoungy:
Have you tested your driver with MySQL? If so, what did you do for the test? Otherwise, I want you to test VerdictDB with MySQL using this driver.

henryqin1997:
@dongyoungy Sure, doing so. What should be the parameters for Mysql such as url, port and schema?  And where can I found makefile for it?

dongyoungy:
I want you to install and run your own MySQL instance. Maybe you can do the following for the test (this is only an example): 
1. Install and run your own MySQL instance.
2. Load TPC-H data
3. Pick one or two TPC-H query
4. Create samples for tables that are used in the query
5. Run query with/without samples to see VerdictDB works and compare the result.

henryqin1997:
Was busy with exam... Now back online. 
@dongyoungy 
1.How can I load TPC-H data(where are them)?
2.I should test queries in /test, right?

PR #137, opened 2018-04-13
explicitly specify dependency of slf4j in pom.xml

dongyoungy:
For issue #133


dongyoungy:
I removed all other slf4j dependencies and added explicit slf4j dependency in root pom.xml. This ensures that we have a single binding of slf4j. A user can specify the version of slf4j by setting sfl4j.version variable while compiling with maven. (e.g., -Dslf4j.version=1.7.5)

PR #146, opened 2018-04-26
Bug fix and approx join improvement

dongyoungy:
I am creating a separate pull request for these changes instead of directly committing to master as they contain changes that are substantial:

1. In ApproxAggregatedRelation, now all aggregated columns are also included in the group-bys. This seems unnecessary, but it is required if ApproxAggregatedRelations is inside a subquery. An example of such case is tpc-ds's [q49](https://github.com/hortonworks/hive-testbench/blob/hdp3/sample-queries-tpcds/query49.sql).
2. Improved the performance of areMatchingUniverseSamples() and sampleType() in ApproxJoinRelation by caching their results. These methods can be super-expensive when there is a huge join, which results in a heavily-nested ApproxJoinRelations.
3. Fixed the logic of extractJoinCondition() in CompCond to find source relation with table name of the column.


pyongjoo:
Sure. I think having an offline discussion will be helpful.

dongyoungy:
The example rewritten query is as follows:

```sql
SELECT * FROM (SELECT ws.`ws_item_sk` AS `item`, (((sum((wr.`wr_return_quantity` / ws.`verdict_vprob`)) / count(*)) * sum(count(*)) OVER (partition by ws.`ws_item_sk`)) / ((sum((ws.`ws_quantity` / ws.`verdict_vprob`)) / count(*)) * sum(count(*)) OVER (partition by ws.`ws_item_sk`))) AS `return_ratio`, (((sum((wr.`wr_return_amt` / ws.`verdict_vprob`)) / count(*)) * sum(count(*)) OVER (partition by ws.`ws_item_sk`)) / ((sum((ws.`ws_net_paid` / ws.`verdict_vprob`)) / count(*)) * sum(count(*)) OVER (partition by ws.`ws_item_sk`))) AS `currency_ratio`, ws.`verdict_vpart` AS `verdict_vpart`, count(*) AS `__vpsize`, avg(0.0) AS `verdict_vprob` FROM tpcds_bin_partitioned_orc_100_verdict.vs_web_sales_uv_0_0100_ws_item_skws_order_number AS ws INNER JOIN tpcds_bin_partitioned_orc_100.web_returns AS wr ON ws.`ws_order_number` = wr.`wr_order_number` AND ws.`ws_item_sk` = wr.`wr_item_sk` INNER JOIN tpcds_bin_partitioned_orc_100.date_dim AS vt8 ON ws.`ws_sold_date_sk` = vt8.`d_date_sk` WHERE (((((wr.`wr_return_amt` > 10000) AND (ws.`ws_net_profit` > 1)) AND (ws.`ws_net_paid` > 0)) AND (ws.`ws_quantity` > 0)) AND (vt8.`d_year` = 2000)) AND (vt8.`d_moy` = 12) GROUP BY ws.`ws_item_sk`, ws.`verdict_vpart`) AS in_web GROUP BY in_web.`item`;
```

PR #168, opened 2018-07-25
Joezhong add show query

Beastjoe:
Add support for describe and show query;

Change VerdictResultStream and VerdictSingleResult to abstract class


src/main/java/org/verdictdb/coordinator/ExecutionContext.java
```diff
@@ -75,20 +85,44 @@ public VerdictResultStream streamsql(String query) throws VerdictDBException {
...
       return stream;
     } else if (queryType.equals(QueryType.scrambling)) {
       ScramblingCoordinator coordinator = new ScramblingCoordinator(context.getCopiedConnection());
       return null;
     } else if (queryType.equals(QueryType.set_default_schema)) {
       updateDefaultSchemaFromQuery(query);
       return null;
-
+    } else if (queryType.equals(QueryType.show_databases)) {
+      return generateShowSchemaResultFromQuery();
```
pyongjoo: Can you assign the return value to a variable; then return the variable? I have found it is easier to debug later.

Please apply the similar approach to the other lines below.

src/main/java/org/verdictdb/coordinator/VerdictResultStream.java
```diff
@@ -16,47 +16,25 @@
...
-  public VerdictResultStream(ExecutionResultReader reader, ExecutionContext execContext) {
-    this.reader = reader;
-    this.execContext = execContext;
-  }
-
   // TODO
-  public VerdictResultStream create(VerdictSingleResult singleResult) {
-    return null;
-  }
+  public abstract VerdictResultStream create(VerdictSingleResult singleResult);
```
pyongjoo: Please change this to a factory method that basically constructs an appropriate regular class and returns it.

pyongjoo: Please add a similar factory method for the ExecutionResultReader.

src/main/java/org/verdictdb/coordinator/VerdictResultStreamFromExecutionResultReader.java
```diff
@@ -0,0 +1,51 @@
...
+  ExecutionContext execContext;
+
+  public VerdictResultStreamFromExecutionResultReader(ExecutionResultReader reader, ExecutionContext execContext) {
+    super();
+    this.reader = reader;
+    this.execContext = execContext;
+  }
+
+  // TODO
+  public VerdictResultStreamFromExecutionResultReader create(VerdictSingleResult singleResult) {
```
pyongjoo: Please just remove this method.

src/main/java/org/verdictdb/coordinator/VerdictResultStreamFromExecutionResultReader.java
```diff
@@ -0,0 +1,51 @@
...
+  public Iterator<VerdictSingleResult> iterator() {
+    return this;
+  }
+
+  @Override
+  public void remove() {
+
+  }
+
+  public void close() {
```
pyongjoo: Please annotate with "@Override".

src/main/java/org/verdictdb/coordinator/VerdictResultStream.java
```diff
@@ -16,47 +16,25 @@
 
 package org.verdictdb.coordinator;
 
-import org.verdictdb.connection.DbmsQueryResult;
-import org.verdictdb.core.resulthandler.ExecutionResultReader;
-
 import java.util.Iterator;
 
-public class VerdictResultStream
+public abstract class VerdictResultStream
```
pyongjoo: Would there be much change needed if we make this class an interface? Otherwise, I feel like interface may make more sense. What do you think?

src/main/java/org/verdictdb/coordinator/VerdictSingleResult.java
```diff
@@ -16,147 +16,36 @@
...
-import org.verdictdb.connection.DbmsQueryResult;
 import org.verdictdb.connection.DbmsQueryResultMetaData;
 
 /**
  * Represents the result set returned from VerdictDB to the end user.
  *
  * @author Yongjoo Park
  */
-public class VerdictSingleResult extends AttributeValueRetrievalHelper {
+public abstract class VerdictSingleResult extends AttributeValueRetrievalHelper {
```
pyongjoo: I feel like interface makes more sense here too. We can place `extends AttributeValueRetrievalHelper` in individual classes. Let me know what you think.

src/main/java/org/verdictdb/coordinator/VerdictSingleResultFromListData.java
```diff
@@ -0,0 +1,158 @@
...
+      Object o = result.get().get(0);
+      if (o instanceof List) {
+        return ((List) o).size();
+      }
+      else return 1;
+    }
+  }
+
+  @Override
+  public String getColumnName(int index) {
```
pyongjoo: Please create a field that contains column names. The field should be initialized in the constructor using the given parameter.

For example, "show table" will display "table name" as a column name. "describe table" will display "column name" in the first column, and "column type" in the second column. These header names should be set by the caller (i.e., in ExecutionContext).

src/main/java/org/verdictdb/coordinator/VerdictResultStream.java
Beastjoe: I think for now there is no big difference between abstract class and interface.

src/main/java/org/verdictdb/coordinator/VerdictSingleResultFromListData.java
Beastjoe: I think describe table will not only return column name and column type. Shall we also include other columns?

src/main/java/org/verdictdb/coordinator/VerdictSingleResultFromListData.java
```diff
@@ -0,0 +1,158 @@
...
+      else throw new RuntimeException("Unexpected columns");
+    }
+  }
+
+  public int getColumnType(int index) {
+    if (result.isPresent() == false || result.get().isEmpty()) {
+      throw new RuntimeException("An empty result is accessed.");
+    } else {
+      Object o = result.get().get(0);
+      if (o instanceof String) return DataTypeConverter.typeInt("varchar");
```
pyongjoo: Please change this to a boring multi-line statement.

src/test/java/org/verdictdb/coordinator/MySqlShowTableTest.java
```diff
@@ -0,0 +1,126 @@
...
+import org.verdictdb.sqlsyntax.MysqlSyntax;
+
+import java.sql.Connection;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+
+import static org.junit.Assert.assertEquals;
+
+public class MySqlShowTableTest {
```
pyongjoo: Please change the name of this class to "MetaDataStatementsTest", and include the tests for other databases. Please refer to this file for avoiding writing the same code: https://github.com/mozafari/verdictdb/blob/master/src/test/java/org/verdictdb/jdbc41/JdbcQueryDataTypeTestForAllDatabases.java

src/main/java/org/verdictdb/coordinator/VerdictResultStream.java
Beastjoe: Wait. I find VerdictResultStream also implement other interfaces.  And there are some override functions in VerdictResultStream. Is it good to move these functions to its subclass?

pyongjoo: No. An interface can also `extends` other interfaces.

src/main/java/org/verdictdb/coordinator/VerdictSingleResultFromListData.java
pyongjoo: What does "describe table" return in addition to column name and column types? Something like primary key information?

Beastjoe: Yes. At least in MySQL, it will return other columns NULL, KEY, DEFAULT and EXTRA

pyongjoo: Our DbmsConnection does not provide that extra information though. For now, let's just return column names and types. We may need to provide more information in the future, which will require updating our DbmsConnection interface too, however.

src/main/java/org/verdictdb/coordinator/VerdictResultStream.java
Beastjoe: You mean create a factory class like queryExecutionPlanFactory that implement create() method?

pyongjoo: Please ignore my above comment, and just convert this class to an interface. Since the interface cannot include any method body in JDK7, my comment is not applicable.

src/main/java/org/verdictdb/coordinator/VerdictSingleResult.java
Beastjoe: If we change VerdictSingleResult to interface, is it safe to change VerdictResultSet  L32 to VerdictResultSetFromDbmsQueryResult? 

```
public class VerdictResultSet implements ResultSet {

  private VerdictSingleResult queryResult;
```

pyongjoo: What would be the problem if we keep VerdictSingleResult there?

pyongjoo: I think it must be VerdictSingleResult to cover both the one based on DbmsQueryResult and the one based on ListData.

Beastjoe: VerdictResultSet is using VerdictSingleResult to implement get method such as getArray(). So it suppose that VerdictSingleResult  extends AttributeValueRetrievalHelper.

pyongjoo: I see. Let's do this.

```java
public interface VerdictSingleResult { 
  (copy those empty get interface methods from DbmsQueryResult [1])
}

public class VerdictSingleResultFromDbmsQueryResult 
    extends AttributeValueRetrievalHelper { ... }

public class VerdictSingleResultFromListData 
    extends AttributeValueRetrievalHelper { ... }
```

[1] https://github.com/mozafari/verdictdb/blob/master/src/main/java/org/verdictdb/connection/DbmsQueryResult.java

src/test/java/org/verdictdb/coordinator/MySqlShowTableTest.java
Beastjoe: For Postgresql and Redshift, these two databases do not use show or desribe, should I test them? Currently, what ExecutionContext is doing is getting list of data from DbmsConnection.

src/main/java/org/verdictdb/coordinator/VerdictSingleResult.java
Beastjoe: Does that make no relationship between VerdictSingleResult and its subclasses?

src/test/java/org/verdictdb/coordinator/MySqlShowTableTest.java
pyongjoo: We should always test them, but the question is just how to test. I think the following is enough.

1. Create a random schema.
1. Create a table in the schema; the table contains two columns (one string type and one integer type).
1. Asserts that the results of our queries are the table name and the column names, respectively.

src/main/java/org/verdictdb/coordinator/VerdictSingleResult.java
pyongjoo: Oh. It should have been like this:
```java
public class VerdictSingleResultFromDbmsQueryResult 
    implements VerdictSingleResult
    extends AttributeValueRetrievalHelper { ... }

public class VerdictSingleResultFromListData 
    implements VerdictSingleResult
    extends AttributeValueRetrievalHelper { ... }
```

src/main/java/org/verdictdb/coordinator/ExecutionContext.java
```diff
@@ -94,33 +94,60 @@ public VerdictResultStream streamsql(String query) throws VerdictDBException {
...
+      return stream;
     } else {
       throw new VerdictDBTypeException("Unexpected type of query: " + query);
     }
   }
 
 
   private VerdictResultStream generateShowSchemaResultFromQuery() throws VerdictDBException {
-    VerdictSingleResultFromListData result = new VerdictSingleResultFromListData((List)context.getConnection().getSchemas());
+    VerdictSingleResultFromListData result = new VerdictSingleResultFromListData(QueryType.show_databases, (List) context.getConnection().getSchemas());
```
pyongjoo: Let's just pass a list here. For example,
```java
List<String> header = Collections.singletonList("schema");
List<List<String>> rows = Collections.singletonList(
    context.getConnection().getSchemas());
VerdictSingleResultFromListData result = 
    new VerdictSingleResultFromListData(header, rows);
```

Please update `VerdictSingleResultFromListData`'s constructor accordingly.

src/main/java/org/verdictdb/coordinator/VerdictSingleResultFromListData.java
```diff
@@ -2,34 +2,47 @@
...
 
+  List<String> fieldsName = new ArrayList<>();
+
   // used to support wasnull()
   private Object lastValueRead;
 
   int cursor = -1;
 
-  public VerdictSingleResultFromListData(List<Object> result) {
+  public VerdictSingleResultFromListData(ExecutionContext.QueryType type, List<Object> result) {
```
pyongjoo: Please change the first argument type to `List<String>` and the second type to `List<List<Object>>`.

You can add another factory method that gets `List<Object>` in the second argument, assuming that the table has only one column.

src/main/java/org/verdictdb/coordinator/VerdictSingleResultFromListData.java
```diff
@@ -2,34 +2,47 @@
...
     super();
     if (result == null) {
       this.result = Optional.absent();
     } else {
+      if (type.equals(ExecutionContext.QueryType.show_databases)) {
```
pyongjoo: Please move this part to the ExecutionContext. The goal is to make this class agnostic to the caller.

src/test/java/org/verdictdb/coordinator/MetaDataStatementsTest.java
```diff
@@ -0,0 +1,388 @@
...
+    }
+  }
+
+  static {
+    REDSHIFT_HOST = System.getenv("VERDICTDB_TEST_REDSHIFT_ENDPOINT");
+    REDSHIFT_USER = System.getenv("VERDICTDB_TEST_REDSHIFT_USER");
+    REDSHIFT_PASSWORD = System.getenv("VERDICTDB_TEST_REDSHIFT_PASSWORD");
+  }
+
+  private static final String SCHEMA_NAME =
```
pyongjoo: Is this for Postgres? Please make it explicit. Note that postgres and redshift are the only systems that distinguish "database" from "schema".

pyongjoo:
It looks good in general. Please make those few changes and push again. Thanks!

Beastjoe:
I have pushed the modified codes. You can have a look.

Issue #179, opened 2018-07-30
Groupby names; Impala show partitions


pyongjoo:
These are currently issues I see. These are fairly minor bugs that can be easily fixed.

1. There are still failed test cases from the groupby name changes.
2. "show partitions" command of Impala throws an exception when there is no partitions. I think we can just catch the exception and returns an empty list inside our JdbcConnection class.

This is related to https://github.com/mozafari/verdictdb/pull/178

Beastjoe:
I think for impala partition, it has already been implemented.
```java
    List<String> partition = new ArrayList<>();
    DbmsQueryResult queryResult;
    if (syntax instanceof ImpalaSyntax) {
      try {
        queryResult = executeQuery(syntax.getPartitionCommand(schema, table));
        for (int i = 0; i < queryResult.getColumnCount(); i++) {
          String columnName = queryResult.getColumnName(i);
          if (columnName.equals("#rows")) {
            break;
          } else partition.add(columnName);
        }
        return partition;
      } catch (Exception e) {
        return partition;
      }
    }
```

PR #180, opened 2018-07-30
Joezhong fix groupby bug in Spark

Beastjoe:
Related to Issue #179. I have comment out some assert in AsyncAggMultiTierTest because I find the output of case when column can be nondeterministic.

### DIFF
--- a/src/main/java/org/verdictdb/core/querying/ola/AggCombinerExecutionNode.java
+++ b/src/main/java/org/verdictdb/core/querying/ola/AggCombinerExecutionNode.java
@@ -105,7 +105,7 @@ public class AggCombinerExecutionNode extends CreateTableAsSelectNode {
     List<SelectItem> allItems = new ArrayList<>();
     
     // replace the select list
-    List<String> groupAliasNames = new ArrayList<>();
+    List<GroupingAttribute> groupingAttributes = new ArrayList<>();
     for (SelectItem item : rightQuery.getSelectList()) {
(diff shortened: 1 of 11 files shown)


pyongjoo:
Thanks for the fix.

I think the non-deterministic ordering is due to this line https://github.com/mozafari/verdictdb/blob/yongjoo/src/main/java/org/verdictdb/core/querying/ola/AggMeta.java#L89

I used HashMap while I should have used TreeMap. Can you do the following and see these changes fix the non-deterministic ordering problem?

1. Change the line to use TreeMap: https://github.com/mozafari/verdictdb/blob/yongjoo/src/main/java/org/verdictdb/core/querying/ola/AggMeta.java#L89
2. Change this line to use TreeMap as well: https://github.com/mozafari/verdictdb/blob/yongjoo/src/main/java/org/verdictdb/core/querying/ola/TierCombination.java#L19
3. Let TierCombination implement Caparable<TierCombination> while the comparison is performed using tableToTier.keySet().

Just for your understanding, tableToTier contains "(schema, table1) -> 0, (schema, table2) -> 1", so tableToTier.keySet() will be "{(schema, table1), (schema, table2)} in this case.

pyongjoo:
Can you also change this line: https://github.com/mozafari/verdictdb/blob/yongjoo/src/main/java/org/verdictdb/core/querying/ola/AsyncAggExecutionNode.java#L280

It is using HashMap too. If the UnnamedColumn does not work for TreeMap, using List should be fine.

Issue #182, opened 2018-07-31
Debug output to a file?


pyongjoo:
It would be nice if the "mvn test" debug output redirects to a file, and we can download the file from CircleCI in case some errors occur.

If CircleCI does not support, please command and close this issue. Thanks.

PR #184, opened 2018-07-31
Prints test log output to a file

dongyoungy:
This request addresses issue #182.
Each CircleCI job prints its log into a file as well. The printed file gets saved into an artifact, which can be downloaded via CircleCI interface.


pyongjoo:
I was thinking of changing the configuration of logback so that a part of its output (debug or below) is only sent to a file. The raw stdout is already available from CircleCI, using `tee` seems to be redundant.

dongyoungy:
I misunderstood what you asked. It is certainly possible to separate the logs, but wouldn't that actually make it harder to debug a bug if one needs to go back and forth between two logs?

pyongjoo:
The scenario I was imaging is the following.

1. Test time
    1. Console only shows the summary (without lengthy sql expressions).
    1. If some error occurs, we open the log file to see more details of the failed test cases (in this case, just searching in the file will be enough)
1. Regular use cases
    1. Console only shows important information (ERROR or WARN). Maybe INFO too.
    1. If some error occurs, we ask the user to send the log file to us.

Let me know if you have another idea.

dongyoungy:
Just pushed a change that works as following:

- During test: 
     console -> INFO or above
     log file (/tmp/verdictdb-debug.log) -> DEBUG or above
- During normal execution: 
     console -> INFO or above
     logfile (/tmp/verdictdb.log) -> DEBUG or above + rolling log for everyday for 30 days max + 1GB cap

I guess saving a log at /tmp/verdictdb.log can be problematic for some, but I am not sure whether there is a better alternative at this point.

pyongjoo:
I noticed time-out error occurred from CircleCI, which could simply be prevented by adding some System.out.pritln() after each test.

PR #186, opened 2018-07-31
Store scramble metadata into ScrambleMetaStore

dongyoungy:
- ScrambleMeta gets stored into target db (default schema: verdictdbmetadata) when a scrambled table is created with 'create scramble...'
- If a user query contains 'old_table' and its scrambled table exists as 'new_table', RelationStandardizer substitutes 'old_table' with 'new_table' -- there is no extra rewriting logic yet.
- Added a logic so that parser parses "IF NOT EXISTS" for "CREATE SCRAMBLE TABLE" as well, but implementation has been added then removed because redshift had a problem with "CREATE TABLE IF NOT EXISTS ... COMPOUND SORTKEY" statement it currently uses for creating its scrambled tables


src/main/java/org/verdictdb/coordinator/SelectQueryCoordinator.java
```diff
@@ -108,7 +119,8 @@ private SelectQuery standardizeQuery(String query) throws VerdictDBException {
     NonValidatingSQLParser sqlToRelation = new NonValidatingSQLParser();
     SelectQuery relation = (SelectQuery) sqlToRelation.toRelation(query);
     MetaDataProvider metaData = createMetaDataFor(relation);
-    RelationStandardizer gen = new RelationStandardizer(metaData);
+    ScrambleMetaStore metaStore = new ScrambleMetaStore(conn);
+    RelationStandardizer gen = new RelationStandardizer(metaData, metaStore);
```
pyongjoo: I think despite some redundancy of the code, it would be much cleaner to separate `RelationStandardizer` and the scramble table replacement logic.

dongyoungy: I will create a separate class for a substitute logic and let you know

pyongjoo:
After this pull request is complete, I will ship a new version to the maven central.

dongyoungy:
I have added a separate ScrambleTableReplace class.

PR #188, opened 2018-08-02
All tests enabled.

pyongjoo:


src/test/java/org/verdictdb/jdbc41/JdbcTpchQueryForAllDatabasesTest.java
```diff
@@ -50,7 +50,7 @@
   // TODO: Add support for all four databases
   //  private static final String[] targetDatabases = {"mysql", "impala", "redshift", "postgresql"};
 //  private static final String[] targetDatabases = {"mysql", "impala", "redshift"};
-  private static final String[] targetDatabases = { "redshift" };
+  private static final String[] targetDatabases = { "impala" };
```
dongyoungy: I think it should same as line 52 (i.e., "mysql, "impala", "redshift"}

pyongjoo: Let me change the line and push again.

PR #196, opened 2018-08-08
SelectQueryCoordinator always gets the most recent ScrambleMetaSet

dongyoungy:
It gets ScrambleMetaSet from ScrambleMetaStore via its retrieve() method before running select queries.


dongyoungy:
VerdictContext now has a ScrambleMetaSet which gets initialized upon its creation.

Issue #202, opened 2018-08-09
Add support for 'BYPASS' command


dongyoungy:
A statement starting with 'BYPASS' keyword should bypass VerdictDB and it should return the results by sending the statement directly to the backend database (without 'BYPASS' keyword).

Issue #194, opened 2018-08-06
NullPointerException when issue statement not supported/used by VerdictDB


dongyoungy:
For instance, if a user wants to drop a table with a Verdict connection, it will generate a NullPointerException because it cannot find a query type for "drop table":

```
Exception in thread "main" java.lang.NullPointerException
	at org.verdictdb.coordinator.ExecutionContext.streamsql(ExecutionContext.java:93)
	at org.verdictdb.coordinator.ExecutionContext.sql(ExecutionContext.java:78)
	at org.verdictdb.jdbc41.VerdictStatement.execute(VerdictStatement.java:48)
	at VerdictTutorialMain.createScrambleTable(VerdictTutorialMain.java:52)
	at VerdictTutorialMain.main(VerdictTutorialMain.java:36)
```

pyongjoo:
Let's just throw this exception: https://github.com/mozafari/verdictdb/blob/master/src/main/java/org/verdictdb/exception/VerdictDBTypeException.java

dongyoungy:
Hmm... I guess we need to bypass such statements that VerdictDB does not support to the original JDBC connection and return the results as-is in the long run.

PR #207, opened 2018-08-10
 Allow 'bypass' keyword in a SQL, which runs the query directly at the backend database

dongyoungy:
Related to #202 and #194

### DIFF
--- a/src/main/java/org/verdictdb/jdbc41/VerdictStatement.java
+++ b/src/main/java/org/verdictdb/jdbc41/VerdictStatement.java
@@ -17,14 +17,14 @@
 package org.verdictdb.jdbc41;
 
 import org.verdictdb.VerdictContext;
-import org.verdictdb.connection.*;
 import org.verdictdb.coordinator.ExecutionContext;
 import org.verdictdb.coordinator.VerdictSingleResult;
-import org.verdictdb.coordinator.VerdictSingleResultFromDbmsQueryResult;
-import org.verdictdb.exception.VerdictDBDbmsException;
 import org.verdictdb.exception.VerdictDBException;
 
-import java.sql.*;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.SQLFeatureNotSupportedException;
+import java.sql.SQLWarning;
 
 public class VerdictStatement implements java.sql.Statement {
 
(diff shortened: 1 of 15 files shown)


src/main/java/org/verdictdb/jdbc41/VerdictStatement.java
```diff
@@ -42,10 +42,67 @@ public VerdictStatement(VerdictConnection jdbcConn, VerdictContext context) {
...
+   */
+  private String checkBypass(String sql) {
+    if (sql.trim().toLowerCase().startsWith("bypass")) {
+      return sql.trim().substring(6);
+    }
+    return null;
+  }
+
+  private VerdictSingleResult executeAsIs(String sql) throws SQLException, VerdictDBDbmsException {
+    DbmsConnection dbmsConn = context.getConnection();
```
pyongjoo: Do we need this check? I guess a simple "dbmsConn.execute(sql)" could work?

src/main/java/org/verdictdb/jdbc41/VerdictStatement.java
```diff
@@ -42,10 +42,67 @@ public VerdictStatement(VerdictConnection jdbcConn, VerdictContext context) {
     this.executionContext = context.createNewExecutionContext();
   }
 
+  /**
+   * Check whether given sql contains 'bypass' keyword at the beginning
+   *
+   * @param sql original sql
+   * @return without 'bypass' keyword if the original sql begins with it. null otherwise.
+   */
+  private String checkBypass(String sql) {
```
pyongjoo: Can we place this logic inside VerdictContext? The Java applications that directly use VerdictContext would not support BYPASS in this way.

pyongjoo: Sounds good to me.

src/main/java/org/verdictdb/jdbc41/VerdictStatement.java
dongyoungy: Argh. I got it. Git highlighted checkByPass() method for the comment.

I guess you are right. I just wanted to be pedantic and use the actual JDBC statements whenever possible, but it does seem redundant at the same time.

dongyoungy:
Changes as suggested have been pushed.

src/main/java/org/verdictdb/jdbc41/VerdictStatement.java
```diff
@@ -45,7 +45,7 @@ public VerdictStatement(VerdictConnection jdbcConn, VerdictContext context) {
   @Override
   public boolean execute(String sql) throws SQLException {
     try {
-      result = executionContext.sql(sql);
+      result = context.sql(sql);
```
pyongjoo: I wonder if this is an intended behavior. My understanding was Connection:VerdictContext = Statement:ExecutionContext. If this was to handle bypass, I would suggest moving the logic into ExecutionContext, which still makes sense.

pyongjoo: It sounds good in general. But I still wonder how to close ExecutionContext when VerdictStatement.close() is called if VerdictStatement does not have the reference to ExecutionContext.

I think their conceptual relationship could be like, when a new VerdictStatement is created, VerdictConnection requests VerdictContext for a dedicated ExecutionContext, and then VerdictStatement simply uses the assigned ExecutionContext for query processing.

I agree that having a reference upper-level structure isn't good.

dongyoungy: Option 1: We can let VerdictContext to manage all of its ExecutionContext and have VerdictStatement to use its VerdictContext to close its associated ExecutionContext (e.g., using Map)

Option 2: We just have 1->2->4 and 3->4 execution path where every query processing logic goes into 4.

For now, I will use Option 2 as suggested while removing references to upper-level class/structures.

Issue #210, opened 2018-08-10
Adding logging level as an option


pyongjoo:
The ability to pass loglevel for each connection will be useful, e.g., "jdbc:verdict:mysql://host:port?loglevel=warn"

My suggestion is to have these two options:
- [console_]loglevel
- file_loglevel

[console_] is optional.

PR #215, opened 2018-08-13
"loglevel" and "file_loglevel" options change the log levels for console and file logs

dongyoungy:
(Related to #210)

Setting 'loglevel' and 'file_loglevel' in VerdictOption changes console and file log levels respectively.

One caveat here is that while it was possible to separate logs from VerdictDB and other libraries for console logs, I could not separate them for file logs as it does not seem to work as intended when we have two appenders for the same log file. So here is how it works for now:

1. For console logs, the changed log level only applies to logs generated by VerdictDB while fixing the log level ('warn' by default in production logback config) in other libraries included
2. For file logs, the changed log level applies to all logs generated by VerdictDB and other libraries as it has to change the log level of root logger

We can change them to always alter root logger's log level or have a separate log files for logs from VerdictDB and others or keep the current implementation.

### DIFF
--- a/src/main/java/org/verdictdb/commons/VerdictDBLogger.java
+++ b/src/main/java/org/verdictdb/commons/VerdictDBLogger.java
@@ -18,14 +18,21 @@ package org.verdictdb.commons;
 
 import ch.qos.logback.classic.Level;
 import ch.qos.logback.classic.Logger;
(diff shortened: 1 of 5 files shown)


pyongjoo:
Can you also add the description to the tutorial docs?

dongyoungy:
Sure. I will add a section for available options in the tutorial.

PR #228, opened 2018-08-20
Fix a bug with 'HAVING' clause when VerdictDB runs the original query on scrambled tables

dongyoungy:
This pull request is currently a draft, suggesting a possible fix to the problem.

The proposed change is to add whatever expression that is in the HAVING clause as-is into the select list of an inner query in a rewritten query so basically the following query:

```SQL
SELECT A
FROM B
GROUP BY ...
HAVING someexpr
```

gets rewritten as something like:

```SQL
...
(SELECT A, someexpr as having_cond
FROM B) as t1
GROUP BY ...
HAVING t1.having_cond = TRUE
```

### DIFF
--- a/src/main/java/org/verdictdb/core/querying/QueryExecutionPlanFactory.java
+++ b/src/main/java/org/verdictdb/core/querying/QueryExecutionPlanFactory.java
@@ -120,7 +120,7 @@ public class QueryExecutionPlanFactory {
     //    query.addOrderby(query.getOrderby());
     //  if (query.getLimit().isPresent()) selectQuery.addLimit(query.getLimit().get());
 
-    int groupbyCount = 1;
+    int groupbyCount = 0;
     for (GroupingAttribute attr : query.getGroupby()) {
       UnnamedColumn groupByCol = findActualGroupByExpression(attr, query);
       query.addSelectItem(
@@ -137,7 +137,7 @@ public class QueryExecutionPlanFactory {
       query.addSelectItem(havingColumn);
     }
 
-    int orderByCount = 1;
+    int orderByCount = 0;
     for (OrderbyAttribute attribute : query.getOrderby()) {
       GroupingAttribute attr = attribute.getAttribute();
       UnnamedColumn orderByCol = findActualGroupByExpression(attr, query);
--- a/src/main/java/org/verdictdb/core/querying/QueryExecutionPlanFactory.java
+++ b/src/main/java/org/verdictdb/core/querying/QueryExecutionPlanFactory.java
@@ -16,23 +16,14 @@
 
 package org.verdictdb.core.querying;
 
-import java.util.ArrayList;
-import java.util.List;
-
 import org.apache.commons.lang3.tuple.Pair;
 import org.verdictdb.coordinator.QueryContext;
+import org.verdictdb.core.querying.ola.AsyncAggExecutionNode;
 import org.verdictdb.core.scrambling.ScrambleMetaSet;
-import org.verdictdb.core.sqlobject.AbstractRelation;
-import org.verdictdb.core.sqlobject.AliasedColumn;
-import org.verdictdb.core.sqlobject.AsteriskColumn;
-import org.verdictdb.core.sqlobject.BaseColumn;
-import org.verdictdb.core.sqlobject.BaseTable;
-import org.verdictdb.core.sqlobject.ColumnOp;
-import org.verdictdb.core.sqlobject.JoinTable;
-import org.verdictdb.core.sqlobject.SelectItem;
-import org.verdictdb.core.sqlobject.SelectQuery;
-import org.verdictdb.core.sqlobject.SubqueryColumn;
-import org.verdictdb.core.sqlobject.UnnamedColumn;
+import org.verdictdb.core.sqlobject.*;
+
+import java.util.ArrayList;
+import java.util.List;
 
 public class QueryExecutionPlanFactory {
 
@@ -105,8 +96,8 @@ public class QueryExecutionPlanFactory {
         selectAll.createPlaceHolderTable("t");
     SelectQuery selectQuery =
         SelectQuery.create(new AsteriskColumn(), baseAndSubscriptionTicket.getLeft());
-//    selectQuery.addOrderby(query.getOrderby());
-//    if (query.getLimit().isPresent()) selectQuery.addLimit(query.getLimit().get());
+    //    selectQuery.addOrderby(query.getOrderby());
+    //    if (query.getLimit().isPresent()) selectQuery.addLimit(query.getLimit().get());
     selectAll.setSelectQuery(selectQuery);
 
     if (query.isSupportedAggregate()) {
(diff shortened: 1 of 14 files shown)


pyongjoo:
Is it correct that the "groupby" is moved outside the logic? It becomes a completely different aggregation then.

dongyoungy:
No, actually group-by should stay in the inner query. For an outer query, an additional group-by should be there by current logic as `someexpr` is not an aggregated column (please correct me if I am wrong). I am sorry about the confusion.

For instance the following query:
```sql
SELECT COUNT("orders"."o_orderkey") as "cnt"
FROM "DBNAME"."orders" "orders"
GROUP BY o_orderstatus
HAVING avg("orders"."o_totalprice") >
(SELECT avg("orders"."o_totalprice") FROM "DBNAME"."orders" "orders")
```

results in a query like following:

```sql
select sum(verdictdb_internal_tier_consolidated."agg0") as "agg0", verdictdb_internal_tier_consolidated."verdictdb_having_cond" as "verdictdb_having_cond" 
from 
(select 1.0 * verdictdb_internal_before_scaling."agg0" as "agg0", verdictdb_internal_before_scaling."verdictdb_having_cond" as "verdictdb_having_cond", verdictdb_internal_before_scaling."verdictdb_tier_alias_696133_0" as "verdictdb_tier_alias_696133_0" from 
(select count(*) as "agg0", avg(vt1."o_totalprice") > (select avg(vt2."o_totalprice") as "a3" 
from "verdictdb_tpch_query_test_zvuulqat"."orders" as vt2) as "verdictdb_having_cond", vt1."verdictdbtier" as "verdictdb_tier_alias_696133_0" 
from "verdictdb_tpch_query_test_zvuulqat"."orders_scramble" as vt1 where vt1."verdictdbblock" = 0 group by "o_orderstatus", vt1."verdictdbtier") as verdictdb_internal_before_scaling) as verdictdb_internal_tier_consolidated 
group by verdictdb_internal_tier_consolidated."verdictdb_having_cond" 
having verdictdb_internal_tier_consolidated."verdictdb_having_cond" = TRUE
```

dongyoungy:
Just realized that the inner group-by column should be added to the outer group-by column as well for this case (i.e., `group by verdictdb_internal_tier_consolidated."verdictdb_having_cond"` should be `group by verdictdb_internal_tier_consolidated,"o_orderstatus", verdictdb_internal_tier_consolidated."verdictdb_having_cond"`?).

pyongjoo:
I think this approach makes sense.

dongyoungy:
Pushed the changes that basically includes:
* Adding expressions in HAVING, GROUP BY, ORDER BY into the select list of the inner query for async aggregation
* Outer query uses those columns to rewrite HAVING, GROUP BY, ORDER BY
* For now, even if there are duplicates columns in HAVING, GROUP BY, ORDER BY (e.g., GROUP BY and ORDER BY on the same column), the current logic simply adds them separately

pyongjoo:
I realized the current logic will not work properly for sum or count is used inside having. An example is
```
SELECT COUNT("orders"."o_orderkey") as "cnt"
FROM "DBNAME"."orders" "orders"
GROUP BY o_orderstatus
HAVING sum("orders"."o_totalprice") > 10;
```

Then, an unscaled having condition is placed inside the inner query and used even when the aggregate is computed only using a part of the data.

I think a maintainable approach is to use the existing scaling logic to properly scale the aggregates in the having condition as well. Specifically,
1. identify aggregate functions in the having condition
2. insert them as additional select items before AsyncAggNode performs any conversions. However, the alias names must be somehow noted.
3. Let AsyncAggNode handle the scaling of those additional aggregates.
4. In the having condition of AsyncAggNode's select query, replace the aggregate functions with the scaled aliases.

This requires parsing the expression for the having clause, but this seems to be actually simpler to handle the cases including a mix of count, avg, and sum.

dongyoungy:
This took a little longer than I expected, due to handling logics that are slightly different between having, group-by, and order-bys.

As suggested by @pyongjoo,
the current implementation basically adds expressions in having, group-by, and order-by into the select list of inner query, then uses those expressions after scaling in the outer query.

I guess some logics may be redundant, but my priority was to make things work for many different scenarios. I did try to optimize how it generates alias names for temporary expressions, though.

For now, it should pass all current unit tests and run redshift Q1-21 without errors.

I think we should add more unit tests later that actually compare the query results between original query and re-written query with scrambled tables. I do not think we have enough of these kinds of tests.

src/main/java/org/verdictdb/core/querying/QueryExecutionPlanFactory.java
```diff
@@ -126,12 +117,61 @@ private static AggExecutionNode createAggExecutionNodeAndItsDependents(
...
+
+    int groupbyCount = 0;
+    for (GroupingAttribute attr : query.getGroupby()) {
+      UnnamedColumn groupByCol = findActualGroupByExpression(attr, query);
+      query.addSelectItem(
+          new AliasedColumn(
+              groupByCol, AsyncAggExecutionNode.getGroupByAlias() + (groupbyCount++)));
+    }
+
+    // dyoon: when 'HAVING' clause is present, its condition is added as a select item in the inner
```
pyongjoo: Can you clarify what is the inner query here? At this stage, I don't think there are such distinctions?

src/main/java/org/verdictdb/core/querying/ola/AsyncAggExecutionNode.java
```diff
@@ -16,43 +16,22 @@
...
-import org.verdictdb.core.sqlobject.SelectQuery;
-import org.verdictdb.core.sqlobject.SqlConvertible;
-import org.verdictdb.core.sqlobject.UnnamedColumn;
+import org.verdictdb.core.sqlobject.*;
 import org.verdictdb.exception.VerdictDBException;
 import org.verdictdb.exception.VerdictDBValueException;
 
-import com.google.common.base.Optional;
-import com.google.common.collect.ImmutableMap;
+import java.util.*;
```
pyongjoo: Can we make this explicit? util.* is huge.

src/main/java/org/verdictdb/core/querying/QueryExecutionPlanFactory.java
dongyoungy: I guess a more precise description would be that this adds those expressions into the select list of the original query, which then gets placed into the inner query of a later re-written query.

pyongjoo: Thanks. I think putting an example could be helpful because even I barely remember exact conversions now... Probably the same thing will happen to all of us.

dongyoungy: Sure. I will try to add an example as a comment for this one.

src/main/java/org/verdictdb/core/querying/ola/AsyncAggExecutionNode.java
dongyoungy: Sure.

dongyoungy:
I found an issue with handling ORDER BYs... I am working on it now.

src/main/java/org/verdictdb/core/querying/ola/AsyncQueryExecutionPlan.java
```diff
@@ -527,120 +511,160 @@ void addTierColumnsRecursively(
...
       rewriteProjectionNodeToAddTierColumn(block, (ProjectionNode) node);
     }
   }
-  
+
   /**
-   * For example, convert
-   * 1. avg(price)            -> sum(price) as 'agg0', count(price) as 'agg1'
-   * 2. sum(price) / count(*) -> sum(price) as 'agg0', count(*) as 'agg1'
+   * For example, convert 1. avg(price) -> sum(price) as 'agg0', count(price) as 'agg1' 2.
```
pyongjoo: I more like the original comment... I guess this is due to the auto indent thing.

pyongjoo:
Since @dongyoungy mentioned there still is a bug, I won't merge this by myself.

Issue #224, opened 2018-08-18
NullPointer exception thrown when an unsupported JDBC url is passed to VerdictContext


pyongjoo:
Tested version: 0.5.4

java.lang.NullPointerException
	at org.verdictdb.VerdictContext.attemptLoadDriverClass(VerdictContext.java:190)
	at org.verdictdb.VerdictContext.fromConnectionString(VerdictContext.java:160)
	at org.verdictdb.TestVerdictDB.main(TestVerdictDB.java:18)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.codehaus.mojo.exec.ExecJavaMojo$1.run(ExecJavaMojo.java:282)
	at java.lang.Thread.run(Thread.java:748)

PR #232, opened 2018-08-25
Throw appropriate VerdictDBException when unsupported jdbc connection string is entered

dongyoungy:


dongyoungy:
This fixes #224

PR #233, opened 2018-08-26
Modify docs based on the documentation

Beastjoe:
I have updated the documentation based on the feedback. Since I don't have the access to directly to this repo, I haven't let mkdocs to deploy the page.


Beastjoe:
I have pushed the changes based on your comments. Please have a check.

Issue #236, opened 2018-08-27
NULL is passed for a cumulative distribution


pyongjoo:
This error seems to be related to the following pull request: https://github.com/mozafari/verdictdb/pull/229

@Beastjoe should take a look.

PR #238, opened 2018-08-27
Joezhong fix null cumulative distribution

Beastjoe:
Related Issue #236 

The problem is that I have comment out ScramblingMethod.getCumulativeProbabilityDistributionForTier() method in ScramblingNode class. 

However, this method will automatically store the distribution by calling storeCumulativeProbabilityDistribution(). I have copied this part to uniform scrambling's getBlockExprForTier() method. It seems good for fast converge since it will call getCumulativeProbabilityDistributionForTier() in getBlockExprForTier().

### DIFF
--- a/docs/docs/getting_started/install.md
+++ b/docs/docs/getting_started/install.md
@@ -56,10 +56,8 @@ You only need a single jar file. This jar file is compiled with JDK8.
     ```
     to check your java compiler version. The output should look like this
     ```
-    $ java -version
-    openjdk version "1.8.0_171-1-ojdkbuild"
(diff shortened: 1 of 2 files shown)


docs/docs/getting_started/install.md
```diff
@@ -50,6 +50,19 @@ You only need a single jar file. This jar file is compiled with JDK8.
     ```
     Check the `target` directory for the created jar files.
 
+    Since Maven requires JDK. Please make sure you have installed JDK. You can use command
+    ```
+    javac -version
+    ```
+    to check your java compiler version. The output should look like this
+    ```
+    $ java -version
```
pyongjoo: This should be also changed as above.

Issue #242, opened 2018-08-29
Supporting count-distinct


pyongjoo:
We should exploit the property of HyperLogLog.

pyongjoo:
@Beastjoe Can you see if
1. we can make no AsyncAgg is done for count-distinct
2. if the backend database supports approximate count-distinct, we use those native functions in the sql writing stage (SqlSyntax class may be modified somehow). FYI, Redshift, Impala certainly have them; Spark seemed to have one too, but please double-check 

In the future, we can implement our own HLL algorithm as UDF for Hive-like engines. Our UDF should return hash values before averaging, so we can average them on VerdictDB side. The major exception is Redshift, which does not provide an efficient way to define UDF. I read online that Its Python-based UDF is much slower.

Beastjoe:
I am thinking the case when both count-distinct and other aggregate function show up in a query that contain scramble table. Is it appropriate to put count-distinct only in the first individual aggregate node and pass it to the AsyncAggNode?

pyongjoo:
I think that approach will make the count-distinct incorrect. Are they test queries that include both count-distinct and other aggregates in a single select list?

Beastjoe:
Yes, I notice that this method is not good. Test queries do not have this case.

pyongjoo:
Then, what we can do is simply reject the queries including both by including error checks. I suggest placing an additional step (as a separate function) after query standardization to check if the submitted query satisfies the structure we can support.

PR #255, opened 2018-09-08
Joezhong check unsupported syntax

Beastjoe:
Related Issue #242 

isSupportQuery() is added in the SelectQueryCoordinator.process(). If query contains not supported syntax (I only include count distinct), then it will throw VerdictDBException


pyongjoo:
I think my suggestion was not clear. Can you modify the current logic as follows?
1. If a select list contains only regular aggregates (i.e., sum, avg, count, min, and max), we support the query. (same as before)
2. If a select list contains only count-distinct, we do not convert it to asyncAgg; however, we still support it (and processes it). In this case, the sql writer converts the count-distinct to approximate count-distinct function in case the backend database supports it.
3. If a select list contains a combination of item 1 (regular aggregates) and item 2 (count-distinct), we throw an error saying "count-distinct cannot appear along with other aggregates."

src/test/java/org/verdictdb/coordinator/RedshiftTpchSelectQueryCoordinatorTest.java
```diff
@@ -1132,7 +1123,7 @@ public void TpchQuery21Test() {
           }
         }
       }
-      */
+      TpchQuery21Test();
```
pyongjoo: Why is this calling itself again?

src/main/java/org/verdictdb/coordinator/SelectQueryCoordinator.java
```diff
@@ -178,13 +178,23 @@ public void abort() {
    * @return false if the query contain the syntax that is not supported by VerdictDB
    */
   private boolean isSupportQuery(SelectQuery query) {
```
pyongjoo: Let's make this function throw an exception in case the query is not supportable. Let the exception include an appropriate message that explains why this query is not supported. Let's change the function name to "ensureQuerySupport".

src/test/java/org/verdictdb/coordinator/RedshiftTpchSelectQueryCoordinatorTest.java
Beastjoe: Oh, that should be a mistake.

Beastjoe: Anyway, I find query 21 has count distinct in having and count in the select list. I think in this situation, this should not be supported because we will convert having into select list.

Beastjoe:
I have slightly changed the AsyncQueryExecutionPlan.createUnfoldSelectlistWithBasicAgg() method because I find that the count(case..) != count(*). So when rewritting the individual aggregate node,  it will not transfer count(case..) to count(*).

src/main/java/org/verdictdb/coordinator/SelectQueryCoordinator.java
```diff
@@ -169,6 +173,64 @@ public void abort() {
...
+          ((ColumnOp)having).doesColumnOpContainOpType("countdistinct")) {
+        containCountDistinctItem = true;
+        ((ColumnOp) having).replaceAllColumnOpOpType("countdistnct", "approx_countdistinct");
+      }
+      if (having instanceof ColumnOp &&
+          having.isAggregateColumn()) {
+        containAggregatedItem = true;
+      }
+      if (containAggregatedItem && containCountDistinctItem) {
+        throw new VerdictDBException("Count distinct and other aggregation should not both appear in one query");
```
pyongjoo: They can appear in two different select lists, right? Like one in an outer query and the other in an inner query. Otherwise, this check should be modified.

Also, please change the error message to "Count distinct and other aggregate functions cannot appear in the same select list."

pyongjoo: What is "this situation"?

Beastjoe: The situation that aggregate item and count distinct are in the selectlist of same query

pyongjoo: My question was if a count-distinct function and another regular aggregate function (e.g., sum) can appear in two different select lists in the same query.

Beastjoe: I see. I think redshift query 13 is the case you mean. 
```
SELECT COUNT(DISTINCT "customer"."c_custkey") AS "ctd_c_custkey_ok",
  "t0"."__measure__0" AS "calculation_1361031105746580"
FROM "public"."customer" "customer"
  INNER JOIN (
  SELECT COUNT((CASE WHEN ((0 < STRPOS("orders"."o_comment",CAST('special' AS TEXT))) AND (0 < (CASE
  	WHEN STRPOS("orders"."o_comment",CAST('special' AS TEXT)) IS NULL THEN NULL
  	WHEN 0 = STRPOS(SUBSTRING("orders"."o_comment",INT4LARGER(1,CAST(FLOOR(STRPOS("orders"."o_comment",CAST('special' AS TEXT))) AS INTEGER))),CAST('requests' AS TEXT)) THEN 0
  	ELSE (STRPOS(SUBSTRING("orders"."o_comment",INT4LARGER(1,CAST(FLOOR(STRPOS("orders"."o_comment",CAST('special' AS TEXT)))AS INTEGER))),CAST('requests' AS TEXT)) + INT4LARGER(1,CAST(FLOOR(STRPOS("orders"."o_comment",CAST('special' AS TEXT))) AS INTEGER)) - 1)	END))) THEN NULL ELSE "orders"."o_orderkey" END)) AS "__measure__0",
    "customer"."c_custkey" AS "c_custkey"
  FROM "public"."orders" "orders"
    RIGHT JOIN "public"."customer" "customer" ON ("orders"."o_custkey" = "customer"."c_custkey")
  GROUP BY 2
) "t0" ON (("customer"."c_custkey" = "t0"."c_custkey") OR (("customer"."c_custkey" IS NULL) AND ("t0"."c_custkey" IS NULL)))
GROUP BY 2
ORDER BY 2
```
I think as long as the asyncNode of inner query performs right, the accuracy will be gradually improved with increasing block number.

pyongjoo: I know that.  So does the function you implemented work as follows?
1. Pass if a count-distinct function and another regular aggregate function (e.g., sum) appear in **two different select lists in the same query**
2. Throws an error if a count-distinct function and another regular aggregate function (e.g., sum) appear **in the same select list.**

Beastjoe: The current logic is checking the select list (also having) to find whether each item contains either aggregate item or count distinct.  If select list contain both it will throw an error. 

And it will do the same process recursively in the from list.

pyongjoo: Thanks for the explanation, but is the answer to my question yes or no? See that my question starts with "does", not "how".

Beastjoe: Yes, it should have. Additionally, the case like
```
select some_column, count(*) 
from table 
group by some_column 
having count(distinct some_column)
```
will fail.

pyongjoo: I see. Would it be hard to place this ensureQuerySupport function after the process "having expression -> select item" is performed? Then, this logic itself does not have to explicitly check the having clause, which will reduce redundancy. If the conversion process is deep inside the query processing logic, the current approach is fine for now.

FYI, "it should have" means "it should" but "it didn't".

Beastjoe: Sorry for the poor syntax. I think it is doable. The having converting process is in QueryExecutionPlanFactory.createAggExecutionAndItsDependents() method. So we can place the checking method in SelectQueryCoordinator right after QueryExecutionPlan is created(). We can track source to find the subqueries of node and recursively check the subqueries.

pyongjoo: I see, but let's keep the current stage then. I think the supported query checking should be performed before converting a query to execution nodes.

Please just change the error message and push. I'll merge this after.

pyongjoo:
@Beastjoe Can you push again after resolving the conflicts?

Issue #257, opened 2018-09-10
Adding the interface for streamsql()


pyongjoo:
@Beastjoe Can you propose what to add and test?

The things I could think of are:
1. Creating a nice print function for a single (individual) result; we can just follow the format offered by typical databases (e.g., MySQL).
2. Adding test cases for streamsql()
    - Retrieving results
    - Ensuring the correctness of individual results
3. Adding a tutorial

Beastjoe:
I think those are enough. 
For creating a nice print function for results, I remember you once said in the future, the print function should additionally add one more column to print the progress or something. (I cannot remember clearly) 
I find this repo https://github.com/htorun/dbtableprinter can be useful for printing function. It is a java utility class to print resultset in database way.

Beastjoe:
I am a bit confused about when should we identify the query type. So far, we identify the query type inside ExecutionContext.streamsql(), which is called by VerdictContext.streamsql(). However, based on your description, I think some function should be created after VerdictStatement.sql() and before VerdictContext.streamsql()

pyongjoo:
When the regular VerdictContext is used, no query type needs to be recognized. It is the user's job to use either sql() or streamsql(). The same select query should be passed to those methods, while the return types are different.

The situation is different when the jdbc interface is used. Since VerdictStatement is now the caller (or the user) of VerdictContext, the VerdictStatement class should recognize the query type and call different methods. Note that the input to the JDBC interface is only a single string query. To distinguish which interface to use (regular or stream), a query can be optionally prepended by "stream".

The only reason that VerdictContext.sql() cannot process "stream select ..." is because its return type is fixed as VerdictSingleResult.

PR #260, opened 2018-09-13
WIP: add StreamSql() interface

Beastjoe:
Related Issue #257 
So far, I have added the method VerdictStatement.sql() that can handle `stream`select query. To make verdictdb can append returning VerdictSingleResult after ResultSet without blocking user, I create a separate thread to execute the query. A new class VerdictStreamResultSet that extends VerdictResultSet is created in order to append the new single result to the resultSet. 

I think the current behavior is after user issues a stream query, a separate thread will run to execute the query and append all returned singleResult to ResultSet. When user calls next(), verdict will return the next row if there are VerdictSingleResults stored in VerdictStreamResultSet. Otherwise, verdict will be blocked until new VerdictSingleResult returns.

There are still a lot of things to do, including making VerdictStreamResultSet support more ResultSet interface and adding tests.


pyongjoo:
Thanks for the progress updates.

Beastjoe:
I have added test for VerdictStatement.sql() of handling stream query and made some changes. For adding print function, In which class users should call print function? How print function should behave? One possible way I am thinking is like automatically calling next() and printing every row. It will be blocked if no result is returned from stream and start to print once it gets result.

pyongjoo:
Please add print() to VerdictSingleResult.

For other classes, i.e., `VerdictStreamResult` and `verdict.jdbc41.ResultSet`, they should work like below.

```java
VerdictStreamResult result = ((VerdictContext) verdict).streamsql("select ...");
for (VerdictSingleResult single : result) {
  single.print()
}
```

```java
verdict.jdbc41.ResultSet result = ((java.sql.Statement) stmt).executeQuery("stream select ...");
while (result.next()) {   // this will be blocked if the next VerdictSingleResult has not processed yet.
  // the first column will always be the seq # of the underlying VerdictSingleResult
  System.out.println(result.getString(1));   
  // for other columns, no difference from the regular interface
  System.out.println(result.getString(2));   
}
```

src/main/java/org/verdictdb/jdbc41/VerdictStreamResultSetMetaData.java
```diff
@@ -0,0 +1,243 @@
...
+import java.sql.ResultSetMetaData;
+import java.sql.SQLException;
+import java.util.Arrays;
+import java.util.List;
+
+import static java.sql.Types.INTEGER;
+
+/**
+ * Created by: Shucheng Zhong on 9/14/18
+ * ResultSetMetaData type for stream select query when VerdictStatement.sql() is called.
```
pyongjoo: Can you tell me why we need this class separately? If there any problem if we simply use the ResultSetMetaData for the first VerdictSingleResult?

src/main/java/org/verdictdb/jdbc41/VerdictStreamResultSet.java
```diff
@@ -0,0 +1,1362 @@
...
+ * Created by: Shucheng Zhong on 9/12/18
+ * ResultSet type for stream select query when VerdictStatement.sql() is called.
+ * <p>
+ * It contains a blocking queue: queryResults. When user call next(), if no queryResult available,
+ * it will try to take one from queryResults. Blocked if no results are returned from VerdictResultStream.
+ * The first column is verdictStreamSequenceColumn, which should be int class that specify the block sequence number
+ */
+public class VerdictStreamResultSet extends VerdictResultSet {
+
+  private static final String verdictStreamSequenceColumn = "verdict_sequence_number";
```
pyongjoo: Let's just use "seq".

src/main/java/org/verdictdb/jdbc41/VerdictStreamResultSetMetaData.java
Beastjoe: Basically they are the same. The reason why I create this class is I find it is hard to adding the new column 'seq' to the original ResultSetMetaData.

src/main/java/org/verdictdb/jdbc41/VerdictStreamResultSet.java
```diff
@@ -0,0 +1,1362 @@
...
+   *
+   * 2) If the other thread gets the mutex, it will append the query result and set the flag to be true. Then next()
+   * takes the mutex. Since there are still queryResult in queryResults, next() will take it from the list so that
+   * the last queryResult won't be neglected.
+   *
+   * @return
+   * @throws SQLException
+   */
+  @Override
+  public boolean next() throws SQLException {
```
pyongjoo: I think `synchronized(this)` keyword could replace this mutex-based mechanism. Please double-check and leave a comment here.

src/main/java/org/verdictdb/jdbc41/VerdictStreamResultSetMetaData.java
pyongjoo: I see. How about adding a column using the VerdictSingleResult (after making a copy)? If possible, I just want to minimize the code redundancy.

src/main/java/org/verdictdb/core/resulthandler/ExecutionResultReader.java
```diff
@@ -63,4 +63,8 @@ public DbmsQueryResult next() {
   public void remove() {
     throw new UnsupportedOperationException();
   }
+
+  public boolean hasEndofQueueReached() {
```
pyongjoo: How is this different from hasNext() defined above?

src/test/java/org/verdictdb/jdbc41/VerdictStatementStreamSqlTest.java
```diff
@@ -0,0 +1,414 @@
...
+import org.verdictdb.jdbc41.VerdictStatement;
+import org.verdictdb.sqlsyntax.MysqlSyntax;
+
+import java.io.File;
+import java.io.IOException;
+import java.sql.*;
+
+import static org.junit.Assert.assertEquals;
+
+public class VerdictStatementStreamSqlTest {
```
pyongjoo: Can you add a comment to this function? I want to know what this test case aims to achieve (or compare).

src/main/java/org/verdictdb/core/resulthandler/ExecutionResultReader.java
Beastjoe: It seems redundant. I will remove it.

src/main/java/org/verdictdb/jdbc41/VerdictStreamResultSetMetaData.java
Beastjoe: It may be difficult to do that because VerdictSingleResult use DbmsQueryResult to get info and DbmsQueryResult, JdbcQueryResult for example, is created directly from ResultSet. Since the interface of those class do not support add column, I think it may be hard to do it without changing the interface.

Beastjoe:
Should print() function also print the sequence number? Since it is under VerdictSingleResult class, it may require some extra info if the sequence number is also needed to be printed. I just find this https://github.com/htorun/dbtableprinter works and it is an open source software. Please have a check to avoid copyright problem.

pyongjoo:
No need to print the seq numbers for VerdictSingleResult. Thanks for the pointer to the library. I'll check.

pyongjoo:
Since https://github.com/htorun/dbtableprinter is GNU licensed, we should indicate that we are using it. Please create (if not exists) "LICENSE-3RD-PARTY.txt" under the root project directory and include the library name, URL, authors and the exact copy of the license itself.

pyongjoo:
So I guess this pull request addresses? https://github.com/mozafari/verdictdb/issues/257

Can you also add some documentation (for docs.verdictdb.org) regarding this feature?

Beastjoe:
Yes, it is for #257. I will start to work on it.

Beastjoe:
I have updated the docs. Please have a check.

src/main/java/org/verdictdb/jdbc41/VerdictStatement.java
```diff
@@ -35,11 +43,82 @@
...
+     * When the last SingleResult is returned, it will synchronized the isCompleted flag of VerdictStreamResultSet
+     * to block possible call of next(). It will set status of VerdictStreamResultSet to be complete and
+     * append the last SingleResult to VerdictStreamResultSet.
+     */
+    public void run() {
+      while (!resultStream.isCompleted()) {
+        while (resultStream.hasNext()) {
+          VerdictSingleResult singleResult = resultStream.next();
+          if (!resultStream.hasNext()) {
+            synchronized ((Object) resultSet.isCompleted) {
```
pyongjoo: Can you elaborate in the above comment about the purpose of this synchronized block? What could happen without the existence of this block?

src/main/java/org/verdictdb/jdbc41/VerdictStatement.java
```diff
@@ -35,11 +43,82 @@
...
+          } else {
+            resultSet.appendSingleResult(singleResult);
+          }
+        }
+      }
+
+    }
+  }
+
+  private Pair<Boolean, String> getStreamQueryIfExists(String query) {
```
pyongjoo: Can you add a comment to describe what happens when the query does not include the "stream" keyword?

pyongjoo:
Thanks for the update. Please make some necessary changes, which shouldn't be hard.

src/main/java/org/verdictdb/jdbc41/VerdictStatement.java
Beastjoe: I think the reason why I added this block is because VerdictStreamResultSet also check isCompleted flag in next() method. I need to make this block atomic to prevent that VerdictStreamResultSet is unaware that the process has ended after the last result is appended. I will add this to comment.

src/main/java/org/verdictdb/jdbc41/VerdictStatement.java
Beastjoe: It will return false. However, I think I should integrate this part to ExecutionContext.identifyQueryType().

src/main/java/org/verdictdb/jdbc41/VerdictStreamResultSet.java
```diff
@@ -0,0 +1,1329 @@
...
+
+  @Override
+  public Date getDate(String columnLabel) throws SQLException {
+    if (colNameIdx.containsKey(standardizedLabel(columnLabel))) {
+      return queryResult.getDate(standardizedLabel(columnLabel.toLowerCase()));
+    } else throw new SQLException("ColumnLabel does not exist.");
+  }
+
+  @Override
+  public Date getDate(String columnLabel, Calendar cal) throws SQLException {
```
pyongjoo: Can we support this method? I think people will complain in the same way for getTimestamp().

src/main/java/org/verdictdb/jdbc41/VerdictStreamResultSet.java
```diff
@@ -0,0 +1,1329 @@
...
+ * <p>
+ * It contains a blocking queue: queryResults. When user call next(), if no queryResult available,
+ * it will try to take one from queryResults. Blocked if no results are returned from VerdictResultStream.
+ * The first column is verdictStreamSequenceColumn, which should be int class that specify the block sequence number
+ */
+public class VerdictStreamResultSet extends VerdictResultSet {
+
+  private static final String verdictStreamSequenceColumn = "seq";
+
+  boolean isCompleted = false;
```
pyongjoo: Let's change this variable name to `hasReadAllQueryResults`.

src/main/java/org/verdictdb/jdbc41/VerdictStreamResultSet.java
Beastjoe: Yes. I will add it.

src/main/java/org/verdictdb/jdbc41/VerdictStreamResultSet.java
```diff
@@ -0,0 +1,1329 @@
...
+  }
+
+  @Override
+  public void cancelRowUpdates() throws SQLException {
+    throw new SQLFeatureNotSupportedException();
+  }
+
+  private void checkIndex(int index) throws SQLException {
+    if (index < 2) {
+      throw new SQLException("Column index must be a positive integer.");
```
pyongjoo: I understand the role, but the message itself is incorrect since the getBlob(1) will call this function, which will then throw this message.

Let's do something like:
```java
if (index == 1) {
  throw new SQLException("The first column of a streaming query result is the sequence number of each result set; thus, the values can be retrieved only in the types compatible with integer.");
} else if (index < 1) {
  throw new SQLException("A column index must be a positive integer.");
}
```

src/main/java/org/verdictdb/jdbc41/VerdictStreamResultSet.java
```diff
@@ -0,0 +1,1329 @@
...
+    }
+  }
+
+  @Override
+  public void clearWarnings() throws SQLException {
+    throw new SQLFeatureNotSupportedException();
+  }
+
+  @Override
+  public void close() {
```
pyongjoo: I think we should call some method to kill the streaming query processing.

src/main/java/org/verdictdb/jdbc41/VerdictStreamResultSet.java
```diff
@@ -0,0 +1,1345 @@
...
+      // on the first call
+      if (queryResult == null && (!hasReadAllQueryResults || !queryResults.isEmpty())) {
+        queryResult = queryResults.take();
+        lastQueryResultIndex++;
+      }
+      boolean hasMore = queryResult.next();
+      if (hasMore) {
+        rowIndex++;
+        return true;
+      } else if (queryResults.peek() == null && hasReadAllQueryResults) {
```
pyongjoo: Is there any difference between hasReadAllQueryResults and isCompleted? (I know I suggested to change the name of the latter variable.)

Beastjoe: No, isn't isCompleted the same with hasReadAllQueryResults?

pyongjoo: Oh. I see. This line previously used isCompleted.

src/main/java/org/verdictdb/jdbc41/VerdictStreamResultSet.java
Beastjoe: I try to use Thread.interrupt(). I think it can work except the running query will throw VerdictDBDbmsException. Is that acceptable?

pyongjoo: Please refer to ExecutionContext.java:L342 for an example.

Beastjoe:
The codes have been updated.

Issue #265, opened 2018-10-02
Adding drop scramble query


pyongjoo:
Interestingly, it looks like we haven't implemented this... Could you add this? FYI, it should also delete the metadata entry, or add a new entry to mark it removed.

dongyoungy:
I guess we should implement three types of queries here:

1. DROP SCRAMBLE <scramble_table> ON <original_table> -> drops a specific scrambled table
2. DROP ALL SCRAMBLE <original_table> -> drops all scrambled table for an original table
3. SHOW SCRAMBLES -> basically shows `verdictdbmeta` table

There is a problem with current metadata entries, because:
1. we should not use DELETE/UPDATE as many HDFS-based DBMS do not support them.
2. we cannot use existing columns to mark the scrambled table as 'deleted'. For example, we may add a new row with 'added_at' column set to NULL for a deleted scrambled table, but if one generates the same scrambled table afterwards, it is hard to distinguish whether it has been deleted or created most recently.

Therefore, we need to add an additional column (say 'deleted' or something like this) in `verdictdbmeta` to check whether the scrambled table has been deleted or not at `added_at` timestamp. ScrambleMetaStore should only return scrambled table entries that have been added later than most recently deleted entry (if one exists) for each <scrambled_table, original_table> pair. 

What do you think?

pyongjoo:
I think the last column is of TEXT type, which is supposed to contain any metadata in JSON format. Maybe, we can use some special key to indicate that this record is for deletion?

dongyoungy:
Oh, you are right. I guess we can simply set it to NULL or some predefined value like 'DELETED'.

pyongjoo:
I think 'DELETED' is good. NULL is always confusing.

dongyoungy:
Got it.

pyongjoo:
Can you release this version to Maven Central as well as on the releases tab?

dongyoungy:
It has been released as v0.5.6.

pyongjoo:
Thanks. Let me close this issue for now. I'll test it in SquirrelSQL.

PR #266, opened 2018-10-03
Add support for dropping scrambled tables

dongyoungy:
(Ref: #265) 
Adds support for the following three query syntaxes:

1. DROP SCRAMBLE `scrambled_table` ON `original_table` : drops a single scrambled table `scrambled_table`
2. DROP ALL SCRAMBLE `original_table` : drops all scrambled table built for `original_table`
3. SHOW SCRAMBLES : show all scrambled tables (basically `select * from verdictdbmeta`)

This pull request also includes a unit test that tests all three types of queries.

### DIFF
--- a/src/main/java/org/verdictdb/metastore/ScrambleMetaStore.java
+++ b/src/main/java/org/verdictdb/metastore/ScrambleMetaStore.java
@@ -66,6 +66,8 @@ public class ScrambleMetaStore extends VerdictMetaStore {
 
   private static final String DELETED = "DELETED";
 
+  private static final String TIMESTAMP_FORMAT = "yyyy-MM-dd HH:mm:ss.SSS";
+
   private DbmsConnection conn;
 
   private String storeSchema;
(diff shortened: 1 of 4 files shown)


src/main/java/org/verdictdb/metastore/ScrambleMetaStore.java
```diff
@@ -115,6 +125,74 @@ public void addToStore(ScrambleMeta scrambleMeta) throws VerdictDBException {
...
+    insertQuery.setSchemaName(getStoreSchema());
+    insertQuery.setTableName(getMetaStoreTableName());
+    String timeStamp = new SimpleDateFormat(TIMESTAMP_FORMAT).format(new Date());
+    insertQuery.setValues(
+        Arrays.<Object>asList(
+            originalTableSchema,
+            originalTableName,
+            scrambleTableSchema,
+            scrambleTableName,
+            DELETED,
```
pyongjoo: The scramble table seems to have 6 columns, but this insert statement has 7 values.

dongyoungy: it actually has 7 columns like the following:
```
+-----------------+-----------+---------+
| name            | type      | comment |
+-----------------+-----------+---------+
| original_schema | string    |         |
| original_table  | string    |         |
| scramble_schema | string    |         |
| scramble_table  | string    |         |
| scramble_method | string    |         |
| added_at        | timestamp |         |
| data            | string    |         |
+-----------------+-----------+---------+
```

PR #252, opened 2018-09-07
WIP: PrestoSyntax

commercial-hippie:
Busy testing this myself, still a work in progress.. Will update as I go.

PR #282, opened 2018-10-12
Add support for Presto + Hive

dongyoungy:
This merges PR #252 with the latest master.
It currently aims to support Presto+Hive configuration only. 
The range of support for Presto's backend connector could be extended later.

This is still WIP and I will add proper unit tests for Presto similar to those of Redshift before review.


commercial-hippie:
Thanks again lot for helping us out with this!

Here are some notes from some tests I ran earlier today.. On a single table (1146003158 rows), on a standalone cluster.  Not sure if this information helps!

```
Query "select count(1) from test_verdict_poi;" took 131.77500009536743 seconds.
// 15:48:00.454 [Thread-1] INFO  o.v.sqlreader.ScrambleTableReplacer - Automatic table replacement: default.test_verdict_poi -> verdictdbscrambles.test_verdict_poi

Query "select count(1) from test_verdict_poi;" took 135.83699989318848 seconds.
// 16:01:32.101 [Thread-1] INFO  o.v.sqlreader.ScrambleTableReplacer - Automatic table replacement: default.test_verdict_poi -> verdictdbscrambles.test_verdict_poi

Query "bypass select count(1) from test_verdict_poi;" took 14.86300015449524 seconds.

Query "bypass select approx_distinct(advertiser_id) from test_verdict_poi;" took 46.407999992370605 seconds.

Query "bypass select count(1) from test_verdict_poi;" took 12.484999895095825 seconds.

Query "select count(1) from test_verdict_poi;" took 116.48199987411499 seconds.
// 16:06:56.691 [Thread-1] INFO  o.v.sqlreader.ScrambleTableReplacer - Automatic table replacement: default.test_verdict_poi -> verdictdbscrambles.test_verdict_poi
```

pyongjoo:
Regarding the longer latency Mike posted, I can think of two reasons:
1. Metadata retrieval took long. We can first check if we are checking metadata more than necessary.
2. Intermediate table creations took long. I have asked @Beastjoe for shortcircuiting table creations for common cases.

@dongyoungy Can you see if you can replicate the problem to identify the cause?

dongyoungy:
I concluded that it is fine to have separate profiles for jdk7 and jdk8 as we only have such dependencies in `test` scope (i.e. currently it is users' responsibility for choosing right jdk + jdbc libraries). 

So the fix will have the following profiles in the pom.xml:
```xml
        <profile>
            <id>jdk7</id>
            <activation>
                <jdk>1.7</jdk>
            </activation>
            <build>
                <plugins>
                    <plugin>
                        <artifactId>maven-surefire-plugin</artifactId>
                        <configuration>
                            <excludedGroups>org.verdictdb.PrestoTests</excludedGroups>
                        </configuration>
                    </plugin>
                </plugins>
            </build>
        </profile>
        <profile>
            <id>jdk8</id>
            <activation>
                <jdk>1.8</jdk>
            </activation>
            <dependencies>
                <dependency>
                    <groupId>com.facebook.presto</groupId>
                    <artifactId>presto-jdbc</artifactId>
                    <version>0.212</version>
                    <scope>test</scope>
                </dependency>
            </dependencies>
        </profile>
```

and CircleCI configs will have something like:
```yml
  build-jdk7:
  ...
      - run:
          name: Run unit tests
          command: mvn -P jdk7 -B test 2> /dev/null
  ...
  build-jdk8:
  ...
      - run:
          name: Run unit tests
          command: mvn -P jdk8 -B test 2> /dev/null
```

As @pyongjoo suggested, future Presto unit tests will be categorized as `PrestoTests` and only be tested with jdk8 profile.

Action items from now are:
1. Add unit tests for Presto just like other databases
2. Try to replicate the issue that @commercial-hippie raised so that we can debug it later

dongyoungy:
I have added a unit test for simple scrambling + querying similar to Redshift unit test. Please take a look and feel free to give me comments on this PR.

PR #285, opened 2018-10-15
Select AsyncAggExecutionNode

Beastjoe:
So far, I have created class SelectAsyncAggExecutionNode and SelectAggExecutionNode. When creating AsyncQueryExecutionPlan, if the source of SelectAll node is AsyncAggExecutionNode, then instead of creating AsyncAggExecutionNode, it will create SelectAsyncAggExecutionNode with SelectAggExecutionNodes as its sources.

I have also created class HashDbmsQueryResult, which is used as the DbmsQueryResult type returned by SelectAsyncAggExecutionNode. 

It now can work for simple cases(Query without subquery). There are still many problems
1. How to transport DbmsQueryResultMetadata of SelectAggExecutionNode to HashDbmsQueryResult.
2. I am not sure whether there is a better way to get aggregate values from base aggregates. For instance, user want to get avg(price), what we get is sum(price) as agg0 and count(price) as agg1. What I am doing is to get the selectItem, which should be ColumnOp(divide, [agg0, agg1]), so I know agg0 should divide agg1 to obtain avg(price). The problem is when ColumnOp gets more complicated, it may be hard to calculate the aggregate directly in memory. So far, I only consider the case of (+ - * /).
3. Test it to support TPCH queries


pyongjoo:
To finish 1 and 2 in a reasonable amount of time, I think we will need some simplifying assumption.

Also, maybe we can rely on third-party libraries (e.g., [tablesaw](https://github.com/jtablesaw/tablesaw)) for in-memory operations.

Beastjoe:
So far, I have passed the test for MySql, Redshift, PostgreSQL and Impala. Spark somehow cannot use H2 driver. It will throw an error saying "There is no suitable driver for H2". I am trying to fix that. 

Also, if SelectAsyncAggExecutionNode is used, many unit tests should be rewritten (Those are used to test the structure of AsyncAggExecutionPlan)

pyongjoo:
@Beastjoe Could you share the current status?

Beastjoe:
@pyongjoo I think I now fix the problem of Spark. But I encountered strange errors of JDK8 on CircleCI. I can run mvn test locally, but CircleCI will fail giving error likes
`[ERROR] ExecutionException The forked VM terminated without properly saying goodbye. VM crash or System.exit called? `
However, CircleCI can run JDK7. Have you encountered this issue before? Thanks!

pyongjoo:
I think https://github.com/mozafari/verdictdb/pull/288 includes some related fixes regarding JDK8.

Beastjoe:
Thanks! It fixes the problem.

Beastjoe:
@pyongjoo Now, I have modified some testcases because of SelectAsyncAggNode. I think it can work now. You can have a check.

src/main/java/org/verdictdb/connection/InMemoryAggregate.java
```diff
@@ -0,0 +1,557 @@
...
+  private static SelectQueryToSql selectQueryToSql = new SelectQueryToSql(new H2Syntax());
+
+  static Connection conn;
+
+  static {
+    try {
+      Class.forName("org.h2.Driver");
+      conn = DriverManager.getConnection(DB_CONNECTION, "", "");
+    } catch (SQLException e) {
+      e.printStackTrace();
```
pyongjoo: Can we make this fail in case these errors occur?

src/main/java/org/verdictdb/connection/InMemoryAggregate.java
```diff
@@ -0,0 +1,557 @@
...
+    } catch (ClassNotFoundException e) {
+      e.printStackTrace();
+    }
+  }
+
+  // key is the non-aggregate column, value is also HashMap,
+  // where the key is the tier value and the value is aggregate columns.
+  // This object stores the results from SelectAggExecutionNode (scaled using ScaleFactor)
+  private HashMap<List<Object>, HashMap<List<Integer>, List<Double>>> results = new HashMap<>();
```
pyongjoo: I guess this field is no longer used?

src/main/java/org/verdictdb/connection/InMemoryAggregate.java
```diff
@@ -0,0 +1,557 @@
...
+    statement.executeBatch();
+  }
+
+  public static DbmsQueryResult executeQuery(SelectQuery query) throws VerdictDBException, SQLException {
+    String sql = selectQueryToSql.toSql(query).toUpperCase();
+    ResultSet rs = conn.createStatement().executeQuery(sql);
+    return new JdbcQueryResult(rs);
+  }
+
+  public static String combine(String combineTableName, String targetTableName, SelectQuery dependentQuery)
```
pyongjoo: If the first argument is the table that contains the values that have been aggregated thus far, let's use the name `combinedTableName`. Using a verb seems to be confusing.

src/main/java/org/verdictdb/connection/InMemoryAggregate.java
```diff
@@ -0,0 +1,557 @@
...
+  }
+
+  private static void resetSchemaAndTableForCombine(UnnamedColumn column) {
+    List<UnnamedColumn> columns = new ArrayList<>();
+    columns.add(column);
+    while (!columns.isEmpty()) {
+      UnnamedColumn col = columns.get(0);
+      columns.remove(0);
+      if (col instanceof ColumnOp) {
+
```
pyongjoo: What is this empty case?

src/main/java/org/verdictdb/connection/InMemoryAggregate.java
```diff
@@ -0,0 +1,557 @@
...
+    // insert values
+    String sql = "INSERT INTO " + tableName + " ("
+        + columnNames
+        + ") VALUES ("
+        + bindVariables
+        + ")";
+    PreparedStatement statement = conn.prepareStatement(sql);
+    while (dbmsQueryResult.next()) {
+      for (int i = 1; i <= dbmsQueryResult.getColumnCount(); i++)
+        statement.setObject(i, dbmsQueryResult.getValue(i - 1));
```
pyongjoo: Do test cases cover timestamp type as well?

src/main/java/org/verdictdb/connection/InMemoryAggregate.java
```diff
@@ -0,0 +1,557 @@
...
+
+  @Override
+  public void printContent() {
+
+  }
+
+  /**
+   * Comparator Class to sort the rows.
+   */
+  class HashMapComparator implements Comparator<Pair<List<Object>, List<Object>>> {
```
pyongjoo: Please remove this class as well if no longer used.

src/main/java/org/verdictdb/connection/InMemoryAggregate.java
Beastjoe: Yes. It runs tests through TPCH queries.

src/main/java/org/verdictdb/connection/InMemoryAggregate.java
Beastjoe: I think it should add its operands into list.

Beastjoe:
I have made the changes. However, when running test on CI, I find it is possible that the unit test stuck at VerdictStatementStreamSqlTest for jdk7 workflow. It can be solved when running it alone. I have run this test multiple times locally with jdk7 and this issue doesn't happen. I am not sure whether it is the problem of CircleCI.

pyongjoo:
I have no idea about the JDK7 issue, but I can see it's a tricky problem. Maybe we can try a different docker image for JDK7? Let me know if you have any other ideas.

pyongjoo:
@Beastjoe Do you think it's safe to merge this? I don't know the status of JDK7 issue.

Beastjoe:
Let me rerun the tests with different images to see whether this issue can be resolved.

Beastjoe:
I think this issue happens occasionally. I rerun the tests five times, and the tests are all passed. So I think it may be safe to merge.

Issue #294, opened 2018-11-09
FYI, QueryResultAccuracyEstimationFromDifference caused the following IndexOutOfBoundsException before the temporary fix:


pyongjoo:
```
java.lang.IndexOutOfBoundsException: Index: 1, Size: 1
	at java.util.ArrayList.rangeCheck(ArrayList.java:657)
	at java.util.ArrayList.get(ArrayList.java:433)
	at org.verdictdb.coordinator.QueryResultAccuracyEstimatorFromDifference.checkConverge(QueryResultAccuracyEstimatorFromDifference.java:64)
	at org.verdictdb.coordinator.QueryResultAccuracyEstimatorFromDifference.isLastResultAccurate(QueryResultAccuracyEstimatorFromDifference.java:45)
	at org.verdictdb.coordinator.ExecutionContext.sql(ExecutionContext.java:152)
	at org.verdictdb.coordinator.ExecutionContext.sql(ExecutionContext.java:123)
	at org.verdictdb.jdbc41.VerdictStatement.executeQuery(VerdictStatement.java:135)
	at org.verdictdb.jdbc41.PrestoSpatialQueryTest.runSimpleSelectSTContainsTest(PrestoSpatialQueryTest.java:171)
```

_Originally posted by @dongyoungy in https://github.com/mozafari/verdictdb/pull/293#issuecomment-437200212_

pyongjoo:
@Beastjoe Can you take a look at this issue?

PR #295, opened 2018-11-09
Joezhong fix 294

Beastjoe:
Related Issue #294 

This problem occurs when the query is selecting asterisk column when there is no AsyncAggExecution node is involved. In that case, the ConsolidatedExecutionNode cannot compress the asterisk query with other query. The metadata `isAggregate` data only judge the asterisk column is not aggregated, so it only has one element, while other metadata has `columnCount()` elements. In that case, I think in that case we should fill the `isAggregate` field with false value.


src/main/java/org/verdictdb/core/querying/simplifier/ConsolidatedExecutionNode.java
```diff
@@ -234,11 +234,13 @@ public ExecutionInfoToken createToken(DbmsQueryResult result) {
     // DbmsQueryResultMetaData should be false for all columns.
     if (token.containsKey("queryResult")) {
       DbmsQueryResult queryResult = (DbmsQueryResult) token.getValue("queryResult");
-      List<Boolean> isAggregate = new ArrayList<>();
-      for (int i=0;i<queryResult.getColumnCount();i++) {
-        isAggregate.add(false);
+      if (queryResult.getColumnCount() != queryResult.getMetaData().isAggregate.size()) {
+        List<Boolean> isAggregate = new ArrayList<>();
+        for (int i=0;i<queryResult.getColumnCount();i++) {
```
pyongjoo: Please add space around '=', around '<', and after ';'

src/main/java/org/verdictdb/core/querying/simplifier/ConsolidatedExecutionNode.java
```diff
@@ -234,11 +234,13 @@ public ExecutionInfoToken createToken(DbmsQueryResult result) {
     // DbmsQueryResultMetaData should be false for all columns.
     if (token.containsKey("queryResult")) {
       DbmsQueryResult queryResult = (DbmsQueryResult) token.getValue("queryResult");
-      List<Boolean> isAggregate = new ArrayList<>();
-      for (int i=0;i<queryResult.getColumnCount();i++) {
-        isAggregate.add(false);
+      if (queryResult.getColumnCount() != queryResult.getMetaData().isAggregate.size()) {
```
pyongjoo: Can this logic handle the case like, "select col1, * from lineitem"?

src/main/java/org/verdictdb/core/querying/simplifier/ConsolidatedExecutionNode.java
```diff
@@ -234,11 +234,13 @@ public ExecutionInfoToken createToken(DbmsQueryResult result) {
     // DbmsQueryResultMetaData should be false for all columns.
```
pyongjoo: Please correct the grammar in the comment.

src/main/java/org/verdictdb/core/querying/simplifier/ConsolidatedExecutionNode.java
Beastjoe: I haven't thought this case. But I think the answer is yes if the query is "select col1, * from lineitem". I think it might be problematic when the query is "select sum(price), * from lineitem group by something". I think I should change the logic slightly to fix this.

pyongjoo: It would be nice if the column information (for filling in the metadata for asterisk) can be obtained from the table names in the from list (of the query).

Beastjoe: I think an easier approach is to compare the column information from queryResult metadata with isAggregate field to find what asterisk column contains.

pyongjoo: That sounds good.

Beastjoe:
I have updated the branch. But jdk8 fails for PrestoSpatialQueryTest:
```
Caused by: org.verdictdb.exception.VerdictDBDbmsException: Issued the following query: insert into "verdictdbmeta"."verdictdbmeta" values ('verdictdb_spatial_query_test_qpeohidm', 'spatial_test', 'verdictdb_spatial_query_test_qpeohidm', 'spatial_scramble', 'uniform', timestamp '2018-11-10 03:47:13.267', '{"schemaName":"verdictdb_spatial_query_test_qpeohidm","tableName":"spatial_scramble","originalSchemaName":"verdictdb_spatial_query_test_qpeohidm","originalTableName":"spatial_test","aggregationBlockColumn":"verdictdbblock","aggregationBlockCount":1,"tierColumn":"verdictdbtier","numberOfTiers":1,"method":"uniform","cumulativeDistributions":{"0":[1.0]}}')
Query failed (#20181110_034713_00675_3zuzf): Error moving data files from hdfs://salat2.eecs.umich.edu:8020/tmp/presto-test/f3eacf19-2734-43d8-9fd4-989cb3bd1f01/20181110_034713_00675_3zuzf_dd3a207f-c569-4427-a723-ccd495ceaf9a to final location hdfs://salat2.eecs.umich.edu:8020/user/hive/warehouse/verdictdbmeta.db/verdictdbmeta/20181110_034713_00675_3zuzf_dd3a207f-c569-4427-a723-ccd495ceaf9a
```

I think it is caused by Presto server?

pyongjoo:
Presto server is likely to be the reason. I was removing some schemas but don't know the current status. Let me double-check.

pyongjoo:
Please rebase on master and see if all tests pass.

Beastjoe:
I just rebased on master and all the tests are passed.

src/main/java/org/verdictdb/core/querying/simplifier/ConsolidatedExecutionNode.java
```diff
@@ -229,6 +241,28 @@ public SqlConvertible createQuery(List<ExecutionInfoToken> tokens) throws Verdic
...
+      DbmsQueryResult queryResult = (DbmsQueryResult) token.getValue("queryResult");
+      if (queryResult.getColumnCount() != queryResult.getMetaData().isAggregate.size()) {
+        List<Boolean> isAggregate = new ArrayList<>();
+        for (int i = 0; i < queryResult.getColumnCount(); i++) {
+          isAggregate.add(false);
+        }
+        for (int i = 0; i < queryResult.getMetaData().isAggregate.size(); i++) {
+          if (!selectQueryColumnAlias.get(i).equals(asteriskAlias)) {
+            int idx = 0;
+            while (!selectQueryColumnAlias.get(i).equals(queryResult.getColumnName(idx))) {
```
pyongjoo: Can you add a comment about this line? What is this condition (inside the parenthesis) effectively testing?

pyongjoo:
Thanks. I just added a single comment for the part that was unclear to me.

PR #301, opened 2018-11-13
Supports count-distinct with hash-based scrambling

pyongjoo:
This pull request will add two features:
1. Enable the following scrambling query:
    ```sql
    CREATE SCRAMBLE newschema.newtable FROM oldschema.oldtable
    METHOD hash
    HASHCOLUMN column_name
    ```
2. Then, approximate count-distinct will be available as follows:
    ```sql
    SELECT count(distinct column_name)
    FROM newschema.newtable
    WHERE ...
    ```
    If `oldschema.oldtable` appears in place of `newschema.newtable` in the above query, an automatic table name replacement is performed.


pyongjoo:
Right at the moment, the test is limited to a simple query: https://github.com/mozafari/verdictdb/pull/301/files#diff-21fda0615f9f54a0d033ff3202a0ea72

pyongjoo:
TODO:
1. use `approx_countdistinct` by default for `count(distinct col)`.
1. double-check the support for presto's `approx_distinct` syntax.

PR #311, opened 2018-11-21
Add support for simpler drop scramble statement

### LINKED ISSUE #308: Simpler "drop scramble"
I noticed that its syntax includes the original table; however, since the scramble itself is a unique identifier, I don't we need the original table name.

I think it is good to have an extra checking, but just in case the user simply issues "drop scramble some_schema.scramble_table", we can drop the table.

dongyoungy:
- Resolves #308 
- Add support for "DROP SCRAMBLE `scramble_schema`.`scramble_table`"
- Original scheme and table names will be `N/A` for the deleted scramble when not specified


pyongjoo:
@dongyoungy Can you rebase on master and ensure that all tests pass? 

I am aware that there are some occasional scaling errors, but let's address it later separately.

dongyoungy:
@pyongjoo I merged the master and it seems to pass all the tests. 

I also added the requested feature that it uses default schema name of the current connection when scramble schema is not specified.

src/main/java/org/verdictdb/coordinator/ExecutionContext.java
```diff
@@ -218,42 +229,42 @@ public VerdictSingleResult sql(String query, boolean getResult) throws VerdictDB
...
+    if (stream == null) {
+      return null;
+    }
+    QueryResultAccuracyEstimator accEst =
+        new QueryResultAccuracyEstimatorFromDifference(selectQuery);
+
+    try {
+      while (stream.hasNext()) {
+        VerdictSingleResult rs = stream.next();
+        ;
```
pyongjoo: Here is just an empty semicolon.

dongyoungy:
Note that I am getting few test failures that seem to happen indeterminately. I guess it is the occasional scaling error that you mentioned. I am sharing two of those for your information:

```
queryCountDistinct3Test - org.verdictdb.coordinator.PrestoTpchSelectQueryCoordinatorTest
java.lang.AssertionError: expected:<2606> but was:<2868>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
	at org.verdictdb.coordinator.PrestoTpchSelectQueryCoordinatorTest.queryCountDistinct3Test(PrestoTpchSelectQueryCoordinatorTest.java:194)
```

```
query17Test - org.verdictdb.coordinator.PostgreSqlTpchSelectQueryCoordinatorTest
java.lang.AssertionError: expected:<5371277.948571429> but was:<6409089.335714285>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:553)
	at org.junit.Assert.assertEquals(Assert.java:683)
	at org.verdictdb.coordinator.PostgreSqlTpchSelectQueryCoordinatorTest.query17Test(PostgreSqlTpchSelectQueryCoordinatorTest.java:403)
```

I will merge this PR nonetheless as the above test failures do not stem from the changes in this PR.

src/main/java/org/verdictdb/coordinator/ExecutionContext.java
dongyoungy: I have removed the extra semicolon

Issue #312, opened 2018-11-22
[MySQL] Retaining a primary key in a scramble


pyongjoo:
When creating a scramble, VerdictDB does not keep the primary key information of the original tables. This makes the MySQL engine chooses an inferior join plan when a query involves joins.

Please refer to the following page for the syntax:
https://dev.mysql.com/doc/refman/5.6/en/create-table-select.html

Please add a new method `getPrimaryKey()` in DbmsConnection.java; and use that method for retrieving the primary key information in the scrambling plan.

solangepaz:
I have the same issue when i running queries with join, using MySQL in verdictdb-tutorial.

solangepaz:
the issue is solved ?

pyongjoo:
No, the code may need to be cleaned and tested more.

solangepaz:
Ok, I really wanted to test the join with VerdictDB today.

pyongjoo:
Then, let me quickly test `git fetch origin pull/314/head:joezhong-fix-312` without merging it into master yet.

solangepaz:
Thank you very much

pyongjoo:
According to my tests, the code is not optimized for mysql, so it was even slower. I will request a fix and test again after the fix.

solangepaz:
I saw that there is a new code, but it has not been tested yet.

solangepaz:
A question, does this error also happen when we use VerdictDB in PostgreSQL?

pyongjoo:
I can affect PostgreSQL as well.

However, there is no problem for SQL-on-Hadoop databases since they don't have the primary key.

pyongjoo:
I have confirmed that the code in [this pull request](https://github.com/mozafari/verdictdb/pull/314) reduces the latency for MySQL join queries. For now, the code is not merged until all tests pass. The test errors seem to be due to CircleCI's issues.

solangepaz:
Ok, good :) so the code still can not be used today ?

pyongjoo:
@solangepaz You can still use the code as follows:
```
git pull
git checkout joezhong-fix-312
mvn -DskipTests -DtestPhase=false -DdeployPhase=false clean install
```
And change VerdictDB's version to "0.5.7-SNAPSHOT" in your pom file.

solangepaz:
Does this is valid in the verdictdb tutorial?

pyongjoo:
[...]

PR #314, opened 2018-11-24
Joezhong fix 312

Beastjoe:
Related issue #312 

Add a field "primarykeyColumnName" when creating scramble tables. If scramble tables have primary keys, it will use this field to add primary key.

### DIFF
--- a/src/main/java/org/verdictdb/VerdictContext.java
+++ b/src/main/java/org/verdictdb/VerdictContext.java
@@ -39,6 +39,7 @@ import org.verdictdb.exception.VerdictDBException;
 import org.verdictdb.metastore.CachedScrambleMetaStore;
 import org.verdictdb.metastore.ScrambleMetaStore;
 import org.verdictdb.metastore.VerdictMetaStore;
+import org.verdictdb.sqlsyntax.MysqlSyntax;
 import org.verdictdb.sqlsyntax.SqlSyntax;
 import org.verdictdb.sqlsyntax.SqlSyntaxList;
 
@@ -149,7 +150,7 @@ public class VerdictContext {
     }
     VerdictOption options = new VerdictOption();
     options.parseConnectionString(jdbcConnectionString);
-    if (jdbcConnectionString.split(":")[1].toLowerCase().equals("mysql")) {
+    if (SqlSyntaxList.getSyntaxFromConnectionString(jdbcConnectionString) instanceof MysqlSyntax) {
       return new VerdictContext(JdbcConnection.create(jdbcConnectionString), options);
     } else {
       return new VerdictContext(ConcurrentJdbcConnection.create(jdbcConnectionString), options);
@@ -176,7 +177,7 @@ public class VerdictContext {
     options.parseConnectionString(jdbcConnectionString);
     options.parseProperties(info);
     options.parseConnectionString(jdbcConnectionString);
-    if (jdbcConnectionString.split(":")[1].toLowerCase().equals("mysql")) {
+    if (SqlSyntaxList.getSyntaxFromConnectionString(jdbcConnectionString) instanceof MysqlSyntax) {
       return new VerdictContext(JdbcConnection.create(jdbcConnectionString, info), options);
     } else {
       return new VerdictContext(ConcurrentJdbcConnection.create(jdbcConnectionString, info), options);
@@ -207,7 +208,7 @@ public class VerdictContext {
     info.setProperty("password", password);
     VerdictOption options = new VerdictOption();
     options.parseConnectionString(jdbcConnectionString);
-    if (jdbcConnectionString.split(":")[1].toLowerCase().equals("mysql")) {
+    if (SqlSyntaxList.getSyntaxFromConnectionString(jdbcConnectionString) instanceof MysqlSyntax) {
       return new VerdictContext(JdbcConnection.create(jdbcConnectionString, info), options);
     } else {
       return new VerdictContext(ConcurrentJdbcConnection.create(jdbcConnectionString, info), options);
@@ -218,7 +219,7 @@ public class VerdictContext {
       String jdbcConnectionString, VerdictOption options) throws VerdictDBException {
     attemptLoadDriverClass(jdbcConnectionString);
     options.parseConnectionString(jdbcConnectionString);
-    if (jdbcConnectionString.split(":")[1].toLowerCase().equals("mysql")) {
+    if (SqlSyntaxList.getSyntaxFromConnectionString(jdbcConnectionString) instanceof MysqlSyntax) {
       return new VerdictContext(JdbcConnection.create(jdbcConnectionString), options);
     } else {
       return new VerdictContext(ConcurrentJdbcConnection.create(jdbcConnectionString), options);
@@ -237,7 +238,7 @@ public class VerdictContext {
     info.setProperty("user", user);
     info.setProperty("password", password);
     options.parseConnectionString(jdbcConnectionString);
-    if (jdbcConnectionString.split(":")[1].toLowerCase().equals("mysql")) {
+    if (SqlSyntaxList.getSyntaxFromConnectionString(jdbcConnectionString) instanceof MysqlSyntax) {
       return new VerdictContext(JdbcConnection.create(jdbcConnectionString, info), options);
     } else {
       return new VerdictContext(ConcurrentJdbcConnection.create(jdbcConnectionString, info), options);
--- a/src/main/java/org/verdictdb/VerdictContext.java
+++ b/src/main/java/org/verdictdb/VerdictContext.java
@@ -149,7 +149,11 @@ public class VerdictContext {
     }
     VerdictOption options = new VerdictOption();
     options.parseConnectionString(jdbcConnectionString);
-    return new VerdictContext(ConcurrentJdbcConnection.create(jdbcConnectionString), options);
+    if (jdbcConnectionString.split(":")[1].toLowerCase().equals("mysql")) {
+      return new VerdictContext(JdbcConnection.create(jdbcConnectionString), options);
+    } else {
+      return new VerdictContext(ConcurrentJdbcConnection.create(jdbcConnectionString), options);
+    }
   }
 
   /**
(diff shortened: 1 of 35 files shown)


pyongjoo:
Thanks. Can you also test the latency with the below query?

The expected behavior is this. Let's Q1 be an internal query used by VerdictDB (i.e., with extra predicates on verdictdbblock). Then, issuing "explain Q1" should shows that "eq_ref" is used for joining tables.

```sql
SELECT SUM(l_extendedprice * (1 - l_discount))
FROM customer, orders, lineitem
WHERE c_mktsegment = 'BUILDING'
AND c_custkey = o_custkey
AND l_orderkey = o_orderkey;
```

Beastjoe:
Yes, customer also has "c_custkey" as its primary key. For scramble tables, I think in MySQL,  a primary key must include all columns in the table's partitioning function.

pyongjoo:
For primary key retrieval, I think creating another node at [this line](https://github.com/mozafari/verdictdb/blob/de30546315832b414b741208bb964e3adf5d94f7/src/main/java/org/verdictdb/core/scrambling/ScramblingPlan.java#L83) is the cleaner approach. Of course, if the SqlSyntax does not support getPrimaryKey() (or does not return null), the node should be ignored. If the `scramblingNode` does not subscribe to the node, it is effectively ignored.

pyongjoo:
Can you double-check the relationship between the partitoning key and the primary key? I just found the following query on [mysql's website](https://dev.mysql.com/doc/refman/8.0/en/partitioning-limitations-partitioning-keys-unique-keys.html):

```sql
CREATE TABLE t5 (
    col1 INT NOT NULL,
    col2 DATE NOT NULL,
    col3 INT NOT NULL,
    col4 INT NOT NULL,
    PRIMARY KEY(col1, col2)
)
PARTITION BY HASH(col3)
PARTITIONS 4;
```

pyongjoo:
My bad. The above query I pasted was an example of invalid statements.

pyongjoo:
Then, can you just compare the latency of running the original join query and Verdict's internal query (on some combination of aggblock)?

Beastjoe:
Sorry for the lone delay. TPCH1g take too long to generate scramble tables so I change the size to 100 Mb. For internal query of VerdictDB, it will take about 0.14s. The original one will take 1.17s to finish.

Beastjoe:
@pyongjoo I have made the change. For testing, I am using 100Mb tpch. The original query takes 1.145s to  finish. With verdict's interface, it may takes 5s. (I assume it means let VerdictStatement runs the sql without using stream)

pyongjoo:
Can you test with TPCH1G? On my machine, creating a scramble for lineitem took about 2 mins.

Also, please analyze what's the most time-consuming part so that we can figure out how to address the issue.

Beastjoe:
@pyongjoo I have tested on TPCH1g. On my machine, original query takes 8.377s to complete, while ours might take 10.043s. However, the result of verdictdb varies a lot(sometimes it only take 4s). I think it is due to the implement of in-memory database. 

I find most of time is consumed in executing the selectAggExeuctionNode. Somehow, the performance is related to the value of maxNumberOfRunningNode in ExecutableRunner L198. The current value is 10. If I change it to 5 or smaller, the time can be reduced to 2.693s. You may try on your machine.

pyongjoo:
Thanks for the update. Are those results with JdbcConnection in place of ConcurrentJdbcConnection?

Beastjoe:
Yes. I have replaced ConcurrentJdbcConnection in VerdictContext for MySQL. You can have a check.

pyongjoo:
I see. My expectation was that the queries issued to the same JdbcConnection are serialized by the backend dbms itself so does not impact the performance how many queries we issue. But, maybe it’s not the case. 

Can you see if we can run individual agg nodes one by one for mysql? I think we should somehow alter NodeRunner, but haven’t look into details.

Beastjoe:
I have updated the code. Now, it takes VerdictDB 1.2s to complete the query.

src/main/java/org/verdictdb/VerdictContext.java
```diff
@@ -149,7 +149,11 @@ public static VerdictContext fromConnectionString(String jdbcConnectionString)
     }
     VerdictOption options = new VerdictOption();
     options.parseConnectionString(jdbcConnectionString);
-    return new VerdictContext(ConcurrentJdbcConnection.create(jdbcConnectionString), options);
+    if (jdbcConnectionString.split(":")[1].toLowerCase().equals("mysql")) {
```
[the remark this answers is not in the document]
pyongjoo: Please apply the same changes to the other lines.

src/main/java/org/verdictdb/connection/DbmsConnection.java
```diff
@@ -60,4 +64,32 @@ public DbmsQueryResult execute(SqlConvertible query) throws VerdictDBException {
...
   public abstract DbmsConnection copy() throws VerdictDBDbmsException;
+
+  /**
+   * @return a list of column names of primary key columns. (0-indexed)
+   */
+  public List<String> getPrimaryKey(String schema, String table) throws VerdictDBDbmsException {
+    List<Integer> primaryKeyIndexList = new ArrayList<>();
+    List<String> primaryKeyColumnName = new ArrayList<>();
+    SqlSyntax syntax = getSyntax();
+    if (syntax.getPrimaryKey(schema, table)!=null) {
```
pyongjoo: Please add spaces before and after `!=`.

pyongjoo: Please move this function to JdbcConnection.

Add getPrimaryKey() in MetaDataProvider.

src/main/java/org/verdictdb/connection/DbmsConnection.java
```diff
@@ -60,4 +64,32 @@ public DbmsQueryResult execute(SqlConvertible query) throws VerdictDBException {
...
+      DbmsQueryResult result = execute(syntax.getPrimaryKey(schema, table));
+      if (syntax instanceof MysqlSyntax) {
```
pyongjoo: I don't think we need this syntax check because "getPrimaryKey() != null" is already used.

src/main/java/org/verdictdb/connection/DbmsConnection.java
```diff
@@ -60,4 +64,32 @@ public DbmsQueryResult execute(SqlConvertible query) throws VerdictDBException {
...
+        while (result.next()) {
+          primaryKeyIndexList.add(result.getInt(3) - 1);
+        }
+      }
+      List<String> columns = new ArrayList<>();
+      result = execute(syntax.getColumnsCommand(schema, table));
+      while (result.next()) {
+        columns.add(result.getString(0));
+      }
+      for (int idx:primaryKeyIndexList) {
```
pyongjoo: Please add spaces before and after ":".

src/main/java/org/verdictdb/core/execplan/ExecutableNodeRunner.java
```diff
@@ -192,19 +190,32 @@ public boolean runOnThread() {
     return false;
   }
 
+  private int setMaxNumberOfRunningNode() {
+    if ((conn instanceof JdbcConnection && conn.getSyntax() instanceof MysqlSyntax)
+        || (conn instanceof CachedDbmsConnection
+          && ((CachedDbmsConnection)conn).getOriginalConnection() instanceof JdbcConnection)
+          && ((CachedDbmsConnection)conn).getOriginalConnection().getSyntax() instanceof MysqlSyntax) {
+      // For MySQL, issue query one be one.
```
pyongjoo: "one be one" -> "one by one"

src/main/java/org/verdictdb/core/execplan/ExecutableNodeRunner.java
```diff
@@ -192,19 +190,32 @@ public boolean runOnThread() {
...
+      if (node instanceof SelectAsyncAggExecutionNode) {
```
pyongjoo: I guess we should add "AsyncAggExecutionNode" as well?

src/main/java/org/verdictdb/core/execplan/ExecutableNodeRunner.java
```diff
@@ -192,19 +190,32 @@ public boolean runOnThread() {
     return false;
   }
 
+  private int setMaxNumberOfRunningNode() {
```
pyongjoo: Change method name to "getMaxNumberOfRunningNode"

src/main/java/org/verdictdb/core/sqlobject/CreateTableAsSelectQuery.java
```diff
@@ -61,6 +63,15 @@ public CreateTableAsSelectQuery(CreateScrambledTableQuery query) {
     this.select = query.select;
     this.overwrite = query.overwrite;
     this.ifNotExists = query.ifNotExists;
+    if (query.primaryKeyColumnName==null) {
```
pyongjoo: Please add spaces before and after "==".

src/main/java/org/verdictdb/sqlwriter/CreateTableToSql.java
```diff
@@ -228,6 +223,18 @@ String createAsSelectQueryToSql(CreateTableAsSelectQuery query) throws VerdictDB
       sql.append("using parquet ");
     }
 
+    // set primary key
+    if (syntax instanceof MysqlSyntax && !query.getPrimaryColumns().isEmpty()) {
+      sql.append("(PRIMARY KEY (");
+      for (String column:query.getPrimaryColumns()) {
+        if (query.getPrimaryColumns().indexOf(column)!=query.getPrimaryColumns().size()-1) {
```
pyongjoo: Please add spaces for "!=".

pyongjoo:
It looks good. I just left some minor comments, but I will go ahead and perform some test with this code.

pyongjoo:
I just pushed some to this branch to change the JdbcConnection.abort() logic slightly.

src/main/java/org/verdictdb/connection/ConcurrentJdbcConnection.java
```diff
@@ -133,6 +133,11 @@ public void setDefaultSchema(String schema) throws VerdictDBDbmsException {
     }
   }
 
+  @Override
+  public List<String> getPrimaryKey(String schema, String table) throws VerdictDBDbmsException {
+    return null;
```
pyongjoo: Please just call
`return getNextConnection().getPrimaryKey(schema, table);`

Beastjoe: I have changed this line.

pyongjoo:
I have merged this pull request into https://github.com/mozafari/verdictdb/pull/315

Issue #321, opened 2018-12-02
NullPointerException on simple count query..


commercial-hippie:
Hi!

When I run the following query verdict throws a nullpointer..

```
SELECT count(*) AS "count" FROM poi_table_orc WHERE day >= 16 AND day <= 30 AND place_type = 'golf_course'
```
If I remove the `place_type = 'golf_course'` part of the query then it runs correctly.  I'll test the query tomorrow on the instance that we created for you to run tests on and add more information here. 

```
Error running instance method java.lang.NullPointerException
	at org.verdictdb.coordinator.QueryResultAccuracyEstimatorFromDifference.checkConverge(QueryResultAccuracyEstimatorFromDifference.java:194)
	at org.verdictdb.coordinator.QueryResultAccuracyEstimatorFromDifference.isLastResultAccurate(QueryResultAccuracyEstimatorFromDifference.java:119)
	at org.verdictdb.coordinator.ExecutionContext.sqlSelectQuery(ExecutionContext.java:249)
	at org.verdictdb.coordinator.ExecutionContext.sql(ExecutionContext.java:160)
	at org.verdictdb.jdbc41.VerdictStatement.execute(VerdictStatement.java:109)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
```

Thanks!

pyongjoo:
We have encountered a similar error in our internal tests, but the fix has not been pushed.

@Beastjoe Can you share the current status on this issue?

Beastjoe:
Sorry for the delay. I will fix this tonight.

pyongjoo:
The related fix has been merged to master: https://github.com/mozafari/verdictdb/pull/322

commercial-hippie:
Thank you :+1: :smile:

pyongjoo:
@commercial-hippie Note that if the filtering predicates do not output any rows, VerdictDB will simply process the entire scramble to produce an accurate answer (in terms of relative errors).  However, your business logic might not want this. That is, you may want the execution to finish early if the absolute difference is expected to be small. In this case, please contact me to discuss how to better suit your case.

PR #322, opened 2018-12-03
Fix nullptr exception when aggregate column contains null value

Beastjoe:
Related issue #321.

The problem is in the execution of selectAsyncAggNode, where H2 jdbc driver will return NULL value if no rows are selected for aggregating. I think adding the check inside QueryResultAccuracyEstimatorFromDifference is enough, since DbmsQueryResult will interpret NULL as 0 automatically.

### DIFF
--- a/src/main/java/org/verdictdb/coordinator/QueryResultAccuracyEstimatorFromDifference.java
+++ b/src/main/java/org/verdictdb/coordinator/QueryResultAccuracyEstimatorFromDifference.java
@@ -8,9 +8,7 @@ import java.util.Set;
 
 import org.verdictdb.VerdictSingleResult;
 import org.verdictdb.commons.TypeCasting;
-import org.verdictdb.core.sqlobject.AsteriskColumn;
-import org.verdictdb.core.sqlobject.SelectItem;
-import org.verdictdb.core.sqlobject.SelectQuery;
+import org.verdictdb.core.sqlobject.*;
 
 /**
  * Estimates the difference based on the difference between two consequent result sets.
@@ -151,7 +149,7 @@ public class QueryResultAccuracyEstimatorFromDifference extends QueryResultAccur
         if (nongroupingColumnIndxes.contains(i)) {
 //        if (currentAnswer.getMetaData().isAggregate.get(i)) {
           // if the aggregate value is null value, we just let it to be 0.
-          if (currentAnswer.getValue(i) == null) {
+          if (checkIfQueryCountOnly() && currentAnswer.getValue(i) == null) {
             aggregateValues.add(0);
           } else {
             aggregateValues.add(currentAnswer.getValue(i));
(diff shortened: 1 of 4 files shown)


Beastjoe:
I think currently, the value inside DbmsQueryResult is NULL if sum select no rows. However, when checking converge in AccuracyEstimator, NULL will be treated as 0.  Is this implementation correct?

pyongjoo:
I think the behavior is correct. Can you add unit tests for confirming it?

Beastjoe:
Since we sum up the aggregate values in selectAsyncAggExecutionNode, count() will also return NULL if no rows are selected. Does it matter?

pyongjoo:
Can you explain more details why the sum of counts will be NULL?

The individual counts could be zero when no rows are chosen; thus, the sum of them could also be zero.

Beastjoe:
The reason is when the aggregation query contains **group-by** and no row is selected, it will return empty set. Therefore, DbmsQueryResult returned from SelectAggExecutionNode contains nothing. Then, SelectAggAsyncAggExecutionNode tries to sum up the result, which results in NULL value for count().

pyongjoo:
Hmm. This is interesting and tricky. Let's examine the original select query, and in case a select item includes count but not other aggregates, replace NULL -> 0. Otherwise, I don't see any harm in keeping null values. FYI, the original select query is already there in the accuracy estimator. I included it for determining aggregate columns.

When it comes to the convergence test, the test must fail for both cases (either 0 or NULL). I think zero values naturally make the test fail, so probably need to add a special treatment for NULL.

src/main/java/org/verdictdb/coordinator/QueryResultAccuracyEstimatorFromDifference.java
```diff
@@ -233,4 +215,18 @@ private boolean checkConverge() {
 
     return isValueConverged;
   }
+
+  public boolean checkIfQueryCountOnly() {
+    for (SelectItem sel:this.originalQuery.getSelectList()) {
+      if (sel instanceof AliasedColumn) {
```
pyongjoo: Instead of this, let's use the function similar to https://github.com/mozafari/verdictdb/blob/master/src/main/java/org/verdictdb/core/sqlobject/ColumnOp.java#L419

Please add a function called "isCountDistinctAggregate()" and call it here.

src/test/java/org/verdictdb/VerdictDBSumNullValueTest.java
```diff
@@ -0,0 +1,122 @@
...
+        scrambler.scramble(MYSQL_DATABASE, "orders", MYSQL_DATABASE, "orders_scrambled", "uniform");
+     meta.addScrambleMeta(meta1);
+    meta.addScrambleMeta(meta2);
+    stmt.execute(String.format("drop schema if exists `%s`", options.getVerdictTempSchemaName()));
+    stmt.execute(
+        String.format("create schema if not exists `%s`", options.getVerdictTempSchemaName()));
+  }
+
+  @Test
+  public void test() throws VerdictDBException {
```
pyongjoo: Can you add tests for count and avg as well?

Beastjoe:
I have made the changes. Please have a check.

Issue #320, opened 2018-11-29
UNION ALL not supporte


commercial-hippie:
I saw that there was a [previous task created](https://github.com/mozafari/verdictdb/issues/122) for this but I think that Verdict has been reworked a lot and support for unions was probably lost somewhere along the way.

Here are some tests I ran.. Excuse the formatting it's from a proxy that returns JSON.

I currently have a workaround implemented for our use case but this would be awesome to have in the core. :smile: 

```
-- "SELECT count(distinct mycolumn) FROM table_b_orc"
{
    "results": [
        {
            "cd2": "4.0"
        }
    ]
}

-- "SELECT count(distinct mycolumn) FROM table_a_orc"
{
    "results": [
        {
            "cd2": "6.0"
        }
    ]
}

-- "SELECT count(distinct mycolumn) FROM table_b_orc UNION ALL SELECT count(distinct mycolumn) FROM table_a_orc"
{
    "results": [
        {
            "cd2": "4.0"
        }
    ]
}

-- "BYPASS SELECT count(distinct mycolumn) FROM table_a_orc UNION ALL SELECT count(distinct mycolumn) FROM table_b_orc"
{
    "results": [
        {
            "_col0": "4"
        },
        {
            "_col0": "6"
        }
    ]
}
```

pyongjoo:
@Beastjoe Could you take a look at this issue (add relevant test cases too)? I think this should be a simple logical issue.

Beastjoe:
@pyongjoo Previously, I think we haven't considered the case of set operation inside user query. You can check visitQuery_expression() method in RelationGen class. It may be a bit complex to fix this. One solution is to treat the SetOperationRelation as a SelectQuery, but I am not sure what to do when scramble tables are involved. Do you have any suggestions? Thanks.

pyongjoo:
SetOperationRelation must not be SelectQuery simply because it is not. SetOperationRelation should just be a subclass of AbstractRelation.

Doesn't the antlr grammar already cover the case that a relation can include UNION? Otherwise, we can simply extend it.

Beastjoe:
Yes, the grammer has already covered the case. QueryNode only handles SelectQuery, I think some similar classes need to be created to handle the set operation inside user query.

Besides, I am not sure what should we do when both relations in set operation involves scramble tables, such as 
```
SELECT count(distinct l_suppkey) FROM lineitem UNION ALL SELECT count(distinct s_suppkey) FROM supplier
```
or 
```
[...]

PR #323, opened 2018-12-10
WIP:Joezhong fix 320

Beastjoe:
Related Issue #320 

Sorry for the delay. I tried to solve the case when scramble tables are involved in set operation. Currently, no AsyncAggExecutionNode is created if scramble tables are in set operation. The reason is that VerdictDB will select and group by verditdbtier column. However, in set operation, the number of selected columns should be the same. For instance, 
`select count(something) from (select something from scramble_table UNION select something from normal_table)`
If we try to use AsyncAggExecutionNode here, the query created by the individual AggExecutionNode will become like
`select something, verdictdbtier from scramble_table where verdictdbblock>=0 UNION select something from normal_table`


Beastjoe:
@pyongjoo 
I think now set operation is supported for async aggregation. When scramble tables have different probability distribution, VerdictDB will not create asyncAggExecutionNode.

PR #326, opened 2018-12-21
Create scrambled table that has same partition columns as its original table

### LINKED ISSUE #303: Presto scramble should keep original partitions
As the title suggests, I think the scrambles should keep the original partitions and just prepend the `verdictdbblock` partition.

Currently it doesn't work that way.. But looking at the code I think that's how it was intended to work?

dongyoungy:
- 'CREATE SCRAMBLE' will create a scrambled table that has same partition columns as its original table (resolves #303)

### DIFF
--- a/src/main/java/org/verdictdb/connection/JdbcConnection.java
+++ b/src/main/java/org/verdictdb/connection/JdbcConnection.java
@@ -16,19 +16,6 @@
 
 package org.verdictdb.connection;
 
-import java.lang.reflect.Constructor;
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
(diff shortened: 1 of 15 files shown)


dongyoungy:
For 'partition by list' syntax that we use to implement partitioned tables in Postgres and MySQL, Postgres only supports a single column partition while MySQL can support partition with multiple columns using 'LIST COLUMNS'.

For now, I have made it to throw an exception for Postgres and MySQL when one tries to create a scramble for already-partitioned tables as suggested. This check is done by existing checkIfSupported() method in CreateScrambleQuery. 

We may implement multi-column partitioned scrambles for MySQL in the future, but I do not think it is necessary at the moment.

PR #327, opened 2018-12-25
Add support for 'scramble appending' (issue #302)

### LINKED ISSUE #302: Scramble Appending
Will there a be way to append to a scramble in the future?

ie. if we have a table which has data added on a daily basis, do we need to drop the scramble and re-create?

I was thinking we might be able to just copy the query used to create the scramble manually (select from) and do a `INSERT (SELECT FROM WHERE 'new data conditions')`.

Could that work or would it mess with the calculations Verdict does?

Thanks!
Mike

dongyoungy:
(Resolve #302. This PR is based on PR #326)
Added 'INSERT SCRAMBLE...' syntax to support scramble appending. 

The current implementation supports the statement of the following form:
```
INSERT SCRAMBLE <existing_scramble_table_schema>.<existing_scramble_table_name> WHERE <condition>
```

This command will append data into existing scramble using same scrambling method + source data that satisfy the given condition.

As a bonus feature, I also added conditional scramble creation like the following:
```
CREATE SCRAMBLE <scramble_table> FROM <original_table> [WHERE <condition>]
```

The current limitations for appending scramble are as follows:
1. New scramble data is appended ONLY from the original source table --> this is because current verdictdb metadata for a scramble has a mapping to its single source table. We may remove this mapping in the future, which can allow a scramble generated from multiple source tables.
1. Appending scramble will not work on scrambles built prior to this version as they do not have necessary information in their metadata to create same type of scramble on new data.
1. It is users' responsibility to make sure not to append duplicates into an existing scramble.

For current implementation, it is required to store classes implementing ScramblingMethod interface (or ScrambilngMethodBase abstract class) as a part of ScrambleMeta in JSON, thus, necessary changes are made to make them serializable (e.g., removing the use of 'Optional' in FastConvergeScramblingMethod).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -111,7 +111,7 @@ jobs:
     # CircleCI maintains a library of pre-built images
     # documented at https://circleci.com/docs/2.0/circleci-images/
     - image: mysql:5.5
-      command: mysqld --innodb_buffer_pool_size=2G --general_log=1 --log_output=/tmp/mysql_general.log --slow_query_log=1 --slow_query_log_file=/tmp/mysql_slow_query.log
+      command: mysqld --innodb_buffer_pool_size=2G --general_log=1 --general_log_file=/tmp/mysql_general.log --slow_query_log=1 --slow_query_log_file=/tmp/mysql_slow_query.log
       environment:
         MYSQL_DATABASE: test
         MYSQL_ALLOW_EMPTY_PASSWORD: "yes"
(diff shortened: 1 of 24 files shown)


dongyoungy:
I made changes so that:
1. it now uses 'cumulativeDistributions' metadata (i.e., `storedProbDist` in ScrambleMethodBase) to figure out block sizes for existing scrambles from prior versions of VerdictDB.
2. a user can use either "INSERT SCRAMBLE ..." or "APPEND SCRAMBLE ..."

dongyoungy:
Removed (WIP) from the title since I think this PR is ready to be reviewed.

Issue #325, opened 2018-12-20
PostgreSQL in PyVerdict


solangepaz:
Hi,
How can I use it PyVerdict in PostgreSQL database ?

pyongjoo:
Pyverdict does not support Postgres yet. We plan to work on it, but do not have a specific ETA yet.

solangepaz:
Hi, 
I tried to adapt the verdictdb tutorial to use with mysql, but I got this error:

![image](https://user-images.githubusercontent.com/15957217/50743488-8617e380-1210-11e9-8877-2b9cf2183f15.png)

solangepaz:
Ok thank you. But does pyverdict now support postgresql? If yes, it's more simple.

pyongjoo:
Thanks for asking. We are following up with this pull request: https://github.com/mozafari/verdictdb/pull/329
It should be merged into master in a couple of days at most.

solangepaz:
Hi,
In relation to use postgresql with verdictdb in java, i added this line at pom.xml:
![image](https://user-images.githubusercontent.com/15957217/51090618-53bd3780-1776-11e9-8b84-f201d6d15ff5.png)

but i have this error again:
![image](https://user-images.githubusercontent.com/15957217/51090610-443dee80-1776-11e9-8ed9-e0c6639e2cce.png)

solangepaz:
I'm not able to do pull request.

But is it already possible to use pyverdict?

XiaoyangShen:
@solangepaz 
Thank you for your patience. The PyVerdict support for PostgreSQL is ready now; You could pull the newest code and compile the source code (tested on miniconda 3.7):
```bash
cd pyverdict
python setup.py install
```
To use PyVerdict for PostgreSQL, please use the following APIs:
```python
import pyverdict
verdict_conn = pyverdict.postgres(host, port, dbname, user, password)
df = verdict_conn.sql(query) # returns a pandas DataFrame containing results
```
More detailed documentations will be released shortly.

solangepaz:
Thank you. But i have this error, when I try to create a scramble table:

`>>> verdict.sql('create scramble tpch1g.nation_x from tpch1g.nation')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ubuntu/miniconda3/lib/python3.7/site-packages/pyverdict/verdictcontext.py", line 226, in sql
    return self.sql_raw_result(query).to_df()
[...]

PR #329, opened 2019-01-07
Pyverdict Postgresql Support

jrectorb:
This pull request addresses issue #325. It adds a pyverdict interface for verdict using postgresql. In particular, this addition follows the typing conventions of popular python library psycopg2 (see https://github.com/psycopg/psycopg2) and thereby provides interfaces accordingly. Note that this pull request required some changes to java portions of the engine, not just the python parts, due to detailed column type names not being otherwise available from the ResultSet JavaObjects returned by verdict's internals.


.circleci/config.yml
```diff
@@ -364,7 +364,7 @@ jobs:
           keys:
           - verdict-dependencies-{{ checksum "pom.xml" }}
       - run: pwd; ls;
-      - run: cd pyverdict; pip install pytest PyMySQL presto-python-client; python setup.py install;
+      - run: cd pyverdict; pip install pytest psycopg2 python-dateutil PyMySQL presto-python-client; python setup.py install;
```
pyongjoo: Since it looks like pyverdict requires python-dateutil, can you specify it in its dependency list in setup.py? Then, I think it should be automatically installed by "python setup.py install".

pyverdict/pyverdict/postgresconverter.py
```diff
@@ -0,0 +1,227 @@
+import psycopg2.tz
```
pyongjoo: Do we need this? If so, this module should also be in the dependency list of setup.py.

.circleci/config.yml
jrectorb: Fixed.

pyverdict/pyverdict/postgresconverter.py
jrectorb: pscopg2 returns time related results with an object of type FixedOffsetTimezone as the tzinfo attribute.  To ensure equality between our results and psycopg2 I returned the same timezone object as they do.  However, I could change this to instead take a more standard tzinfo value and check for equality differently in the tests.  For now I've added psycopg2 as a dependency in setup.py, though.  I can change this if useful to be handled in equality checking of the tests if useful.

pyongjoo: Let's return something standard e.g., Python's datetime.datetime, rather than the special type. For testing, we can simply compare the content, not the types. Lastly, please remove the pscopg2 from the dependency.

PR #330, opened 2019-01-09
Pyverdict Impala Support

jrectorb:
This pull request adds impala support for pyverdict.  The pull request respects datatypes as in the impyla library (https://github.com/cloudera/impyla), the main python interface for impala.

Note that to properly test the pyverdict impala connector we ought compare the results of queries made through the connector to results provided by impyla.  However, currently impyla does not support python 3.7.  Thus, in testing we create a separate downgraded venv using python 3.6 to run pyverdict impala tests with, then use the original venv with python 3.7 to run all remaining pyverdict tests.


pyverdict/pyverdict/verdictcontext.py
```diff
@@ -62,30 +74,98 @@ def close(self):
...
 
     @classmethod
-    def new_presto_context(cls, host, catalog, user, password=None, port=8081):
+    def new_presto_context(
+        cls,
+        host,
+        catalog,
+        user,
+        password=None,
+        port=8081,
```
pyongjoo: Can you change this port to 8080? This was my mistake.

PR #334, opened 2019-01-13
automatic table replacement logic changed to detect scramble table instead

### LINKED ISSUE #333: Remove automatic table detection for scrambles
We need to remove automatic table detection for scrambles and let users to specify scrambles directly if one wants to use AQP with VerdictDB. This needs to be done because:

1. It is currently very difficult to support automatic table detection properly for complex queries (e.g., ones with nested queries) with multiple scramble types.
1. It provides more freedom to users to specify which scramble to use for a given query. This is possible because we now allow users to view available scrambles (with `SHOW SCRAMBLES` command).

We may go back to automatic table replacement for scrambles once we figure out the proper way to do it, but until then we think we should let users to specify which scramble to use for his/her queries.

dongyoungy:
Resolves #333 

- User needs to specify scramble directly in the query -- VerdictDB no longer automatically replaces an original table with its scramble
- VerdictDB detects the specified scramble applies appropriate AQP logic

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Documentation: https://docs.verdictdb.org
 
 
 
-# Same SQL, Same DB, 100x-200x Faster Analytjjjjics
+# Same SQL, Same DB, 100x-200x Faster Analytics
 
 <p align="center">
 <img src="http://verdictdb.org/image/verdict-for-impala-speedup.png" width="600px" />


dongyoungy:
Updated README.md a little bit so that it does not conflict with what we do currently at least. 
You can check it out [here](https://github.com/mozafari/verdictdb/tree/dyoon-feature-333).

pyongjoo:
Looks good. Thanks.

PR #337, opened 2019-01-16
Joezhong stratified sampling

Beastjoe:
Here is the group size distribution on l_partkey and l_suppkey in tpch1g. The y axis is the number of groups of same size and the x-axis is the group size. I think it can be assumed as normal distribution.
l_partkey
l_suppkey

The scrambling method will first create a temp table to count the group size, such as
`CREATE [temp table] AS SELECT [stratified column], count(*) from [original table] group by [stratified column]`
Then some statistical nodes will calculate the average and standard deviation of group size and count the size of rare group. After that, it will create scramble table using some similar logics used in fast converge method.
Currently, the rare group(tier 0) is defined to be [average group size] - 1.0*[standard deviation]. It supports one stratified column.


pyongjoo:
Can you tell me what are the differences from the existing FastConverge method?

Beastjoe:
The main difference is FastConverge divides group using the value of columns, while Stratified method uses the size of group. For instance, if a column is numeric type, FastConverge will calculate the avg and stddev of this column and set the outlier group to be any value lower or higher than the boundary (calculated using avg(column) and std(column)).

Another difference is FastConverge will apply to every columns that are numeric, while Stratified only applies to the column users specify.

pyongjoo:
I see. Thanks for the clarification.

Beastjoe:
@pyongjoo @dongyoungy 
I have added the implementation of stratified sampling. So far, it will put all rare groups into block 0 and do uniform sampling in the remaining blocks. The syntax is like
`create scramble [scramble table] from [original table] METHOD stratified STRATIFIED BY [columns] RATIO [uniform sampling ratio] LEASTSAMPLINGSIZE [minimum sampling size]`

src/main/antlr4/org/verdictdb/parser/VerdictSQLParser.g4
```diff
@@ -70,9 +70,11 @@ create_scramble_statement
     : CREATE SCRAMBLE (IF NOT EXISTS)? scrambled_table=table_name FROM original_table=table_name
       (WHERE where=search_condition)?
       (METHOD method=scrambling_method_name)?
-      ((HASHCOLUMN | ON) hash_column=column_name)? 
+      ((HASHCOLUMN | ON) hash_column=column_name)?
+      (STRATIFIED BY stratified_column=column_name_list)?
```
dongyoungy: I think it could be better if we keep a single syntax for getting column names.
Maybe we can have `((HASHCOLUMN | ON) hash_column=column_name_list)?` for columns required by a scramble. For hash scrambles, we can just take first column in the list (for now).

src/main/java/org/verdictdb/core/scrambling/StratifiedScramblingMethod.java
```diff
@@ -0,0 +1,455 @@
...
+  private double sampleingRatio = 0.1;
+
+  public static final String MAIN_TABLE_SOURCE_ALIAS_NAME = "t1";
+
+  public static final String GROUP_INFO_TABLE_SOURCE_ALIAS_NAME = "t2";
+
+  private int totalNumberOfblocks = -1;
+
+  private static VerdictDBLogger log =
+      VerdictDBLogger.getLogger(FastConvergeScramblingMethod.class);
```
dongyoungy: You can put `StratifiedScramblingMethod.class` for the logger here

Beastjoe:
@dongyoungy Shouldn't block 0 only contain rows from rare groups? I guess you mean combing block 0 and block 1?
* block0 has at least k*a rows for each group

dongyoungy:
I thought the idea later proposed by Young was something like:
- <tier0, block0> : all rows from rare groups
- <tier1, block0> : rows from other non-rare groups

and all rare groups go tier0 and others continue with tier1?

Beastjoe:
I see. I will change the logic.

Beastjoe:
I have added the tests. For groups that are not rare(#row> k/a), I think we can only guarantee statistically that at least k rows are selected in each block.

dongyoungy:
> pick top `table_size * prob` rows

To correct, I guess it should be `group_size * prob` rows as we are dealing with individual groups in a table.

Beastjoe:
I just realized if we could use row_number() to enforce the row number of tier1 group, we can also enforce the row number of tier 0 group, like 
```
SELECT * 
FROM (SELECT *, row_number() as rn FROM table ORDER BY rand()) tmp
WHERE rn <= K
```
where K is the minimum sampling size. Then, we can avoid putting too much rows in block 0. What do you think?

Beastjoe:
I have fixed all the unit tests. You can have a check. @pyongjoo @dongyoungy

.circleci/config.yml
```diff
@@ -39,7 +39,7 @@ jobs:
     # Specify service dependencies here if necessary
     # CircleCI maintains a library of pre-built images
     # documented at https://circleci.com/docs/2.0/circleci-images/
-    - image: mysql:5.5
+    - image: mysql:8.0
```
dongyoungy: Is this version change intended by Young?

src/main/java/org/verdictdb/sqlsyntax/ImpalaSyntax.java
```diff
@@ -114,6 +114,15 @@ public String randFunction() {
     return String.format("rand(unix_timestamp()+%d)", randomNum);
   }
 
+  @Override
+  public String rowNumberFunction(String partition) {
+    if (partition.equals("")) {
+      return "ROW_NUMBER() OVER ()";
+    } else {
+      return String.format("ROW_NUMBER() OVER (PARTITION BY %s ORDER BY %s)", partition, randFunction());
```
dongyoungy: Why `ORDER BY rand()` only in Impala? Is it randomly ordered for other DBMS by default?

Beastjoe: Yes. At least in the current version, Impala require order by clause.

.circleci/config.yml
Beastjoe: I changed this since MySQL with the version under 8.0 does not support row_number() .

src/main/java/org/verdictdb/sqlsyntax/ImpalaSyntax.java
dongyoungy: So all the other DBMS do not require `ORDER BY` and they randomly order within the specified partitions by default? I just want to make sure.

dongyoungy:
@pyongjoo I think this PR is almost ready to be merged, except it now requires MySQL 8.0 instead of 5.5 due to its use of row_number() func. 

Please take a look briefly and make comments on this PR if necessary.

src/main/java/org/verdictdb/sqlsyntax/ImpalaSyntax.java
Beastjoe: Yes, I think Sqlite need ORDER BY as well. But I just find Sqlite syntax hasn't been implemented yet.

Beastjoe:
@dongyoungy @pyongjoo I have changed the logic of stratified sampling based on Yongjoo's comment. So far, for groups that are not rare, we divides them into multiple tiers (10 tiers in default) and do uniform sampling. For instance, if user specify at least 100 rows for each group in each block, and sampling ratio is 0.1, block size is 100, then
- For groups of which #row < 1000, we put them in first block
- For groups of which #row < 10000, we also put them in first block 
- For groups of which 10000 < #row < 20000, we uniformly sample within first 10 blocks
- For groups of which 20000 < #row < 30000, we uniformly sample within first 20 blocks
...
- For groups of which 90000 < #row < 100000, we uniformly sample within first 90 blocks
- For groups of which #row > 100000, we simply do uniform sampling

Yongjoo said it is okay if we can guarantee the expectation rows of each group in each block is at least k. So, row_number() is not used.

Issue #346, opened 2019-02-11
Dual table names in tutorial


Chen-Huanyi:
In section
[Apache Spark. Create table and load data](https://docs.verdictdb.org/tutorial/tpch/#create-table-and-load-data)
The SQLs are not correct, the last line of each query should be as:

 LOCATION '/tmp/tpch1g/nation/**~~nation~~**';

pyongjoo:
@Chen-Huanyi Thanks for telling us the problem.

@dongyoungy Can you assign this someone to take a look? Thanks.

dongyoungy:
Will be fixed by PR #347

PR #345, opened 2019-02-07
Docs

ZJCSJTU:
added subsection 'TPC-H Application'.

PR #347, opened 2019-02-11
Documentation updates (same as #345)

dongyoungy:
Basically same as #345, but with a few minor fixes.


dongyoungy:
This will resolve #346

Issue #351, opened 2019-02-25
Creating empty scramble should throw exception


commercial-hippie:
When a scramble is created on a empty data set I think it should throw an error to avoid broken scrambles.  (I tried to append to the scramble later).

Here is a example I got from a scramble on a empty table:

```
{"schemaName":"verdict_scrambles","tableName":"my_table","originalSchemaName":"default","originalTableName":"my_table","aggregationBlockColumn":"verdictdbblock","aggregationBlockCount":1,"tierColumn":"verdictdbtier","numberOfTiers":1,"method":"UNIFORM","scramblingMethod":{"blockSize":1000000,"type":"uniform","maxBlockCount":100,"relativeSize":0.1,"storedProbDist":{"0":["Infinity"]},"MAIN_TABLE_SOURCE_ALIAS":"t","totalNumberOfblocks":0,"actualNumberOfBlocks":1,"mainTableAlias":"t","tierCount":1,"blockCount":0,"actualBlockCount":1},"hashColumn":null,"cumulativeDistributions":{"0":["Infinity"]}}
```

```
"cumulativeDistributions":{"0":["Infinity"]}
"storedProbDist":{"0":["Infinity"]}
```

pyongjoo:
I'm sorry that no one has been following up with this issue.

@dongyoungy Can you take a look into this? The fix seems to be pretty straightforward.

dongyoungy:
I have already asked Jiacheng to look at an efficient way of determining an empty table to solve this issue.

pyongjoo:
I see. That's great.

commercial-hippie:
No worries, thank you!

Beastjoe:
Hi, I have created a pull request to solve this problem. In the process of creating scramble, it will throw a VerdictDBException when empty table is detected.

dongyoungy:
@commercial-hippie FYI, the fix for this issue has been merged into master.

PR #353, opened 2019-03-07
throw exception when creating scramble on empty table

Beastjoe:
Related Issue #351 
VerdictDB will throw exception if empty table is used to create scramble. I do this by checking totalBlockSize in ScrambleNode.createQuery(). If totalBlockSize==0, it is equivalent to say the table size is 0.


dongyoungy:
I first thought we had to use different method for determining an empty table as opposed checking the size of the table, but I guess if it is an empty table, `SELECT COUNT(*)` will run quickly anyway so it seems fine to do it this way.

PR #358, opened 2019-03-26
auto tpch tests

ZJCSJTU:


pyongjoo:
Please remove the directory 2.8.0_rc2 (and files under the directory).

Is gendata the script you wrote?

Place sql queries in regular text files. Let your test script load sql from those files and run tests. Then, a single file should be enough for running all tests.

ZJCSJTU:
The directory 2.8.0_rc2 includes an executable (compiled on my machine) and one other file to generate new data. But I can rename it  a better name.
Yes, I wrote gendata. But the sql queries depend on the size of tpch test given by the user. For example, I create database tpchXg for each X-size test, so one can have different-size tests on one machine without removing one before doing another. 
I can put gendata in tpchtest, so only tpchtest will be needed to run tests.

pyongjoo:
Compiled C code is machine dependent. Git is for source code not for compiled executable.

Let your script accept an argument, which indicates the root directory of the data folders. Make your script work under the assumption that the data files are stored as follows:
(arg directory)
- lineitem
    - lineitem.tbl
- orders
    - orders.tbl
- and so on

ZJCSJTU:
Should I change it to be running test given the existing data (instead of generating data of size given by the user)? Because otherwise those tpch source codes need to be there.

pyongjoo:
Yes, please let the script run given the existing data.

ZJCSJTU:
I have changed it in that way.

pyongjoo:
Please do this as well.

"Place sql queries in regular text files. Let your test script load sql from those files and run tests. Then, a single file should be enough for running all tests."

PR #366, opened 2019-05-15
Disable unit tests that use private resources

dongyoungy:
This is an attempt to enable automatic unit tests on external pull requests via CircleCI.
This PR:
1) disable unit tests using our private resources via `@Ignore` annotation
2) comment out `git pull` on our private resources in the CircleCI configuration

Please note that:
1) Currently our Redshift test instance is down, so CircleCI tests will fail
2) There has been issue with CircleCI image that we use for our jdk7 tests: https://discuss.circleci.com/t/apt-get-update-return-404/29237/8
I have included a temporary fix to it, but we may need to use a different image for a proper fix.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@
 general:
   branches:
     ignore:
-    - gh-pages
+      - gh-pages
 
 version: 2
 jobs:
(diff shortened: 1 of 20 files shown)


pyongjoo:
I think it looks good. Let me disable the private repo itself and let you know. In this way, we can fundamentally guarantee no access to the repo.

pyongjoo:
@dongyoungy Currently, the errors are due to our stopped Redshift cluster.  Can you also simply deactivate those unit tests?

pyongjoo:
@dongyoungy Can you fix the error and merge this to the master?  Not urgent, but I would like to base my branch on this.

dongyoungy:
I have further disabled unit tests with redshift since all unit tests get passed, I am merging this PR.

PR #378, opened 2019-06-04
Run Presto unit tests locally using a docker image

dongyoungy:
This PR makes Presto unit tests to run locally on `starburstdata/presto` docker image in build-jdk8.

The original plan was to use `tpch.tiny` schema (because even `tpch.sf1` was too big) for tpch queries, but it turned out that it was also too big for our CircleCI instances as Presto docker container died with exit code 137, meaning it was likely died due to OOM -- as it now uses `memory` catalog, creating scrambles in the memory during the tests --, while it worked fine locally on my machine.

Therefore, I had to go back to the existing way of populating tpch data manually as before in `memory` catalog. I had to change some parameters in CircleCI config for build-jdk8 as the container still went OOM and died during the test. Now it seems to work except that it fails on some Impala tests, which has nothing to do with this PR and it just looks like our Impala test instance is not operational currently.

Overall, it took some time for me to implement the 'original plan' that I had, but I had to go back to previous method, which made many of changes that I had made unnecessary :(

Nonetheless, the changes include:

- Fix build-jdk8 config for memory tweaks and also running presto docker container
- Support for optional **catalog.schema.table** format for table reference
- Add `PrestoMemorySyntax` for unit tests: 'memory' catalog does not support a few syntax that works for other catalogs (e.g., parititon_by)
- Add 'SetupPrestoInMemory' and other related methods to create tpch test data in 'memory' catalog
- New tpch query templates for `tpch.tiny` in the `presto_new` directory (ended up not using them though)
- Fix few queries in previous presto tpch queries (basically removed date function)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -183,32 +183,33 @@ jobs:
           POSTGRES_DB: test
           POSTGRES_USER: postgres
           POSTGRES_PASSWORD: ""
+      - image: starburstdata/presto:latest
     working_directory: ~/repo
     environment:
       # Customize the JVM maximum heap limit
       MAVEN_OPTS: -Xmx3200m
     steps:
       - run: apt-get update; apt-get install -y git wget
+      #      - run:
+      #          name: Install Docker client
+      #          command: |
+      #            set -x
+      #            VER="18.09.6"
+      #            curl -L -o /tmp/docker-$VER.tgz https://download.docker.com/linux/static/stable/x86_64/docker-$VER.tgz
+      #            tar -xz -C /tmp -f /tmp/docker-$VER.tgz
+      #            mv /tmp/docker/* /usr/bin
       - run:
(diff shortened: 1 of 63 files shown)


pyongjoo:
This is great. Thanks!

Can you also disable Impala tests for now? I think we can focus on Presto (and possibly Spark).

pyongjoo:
@dongyoungy I see many changes for supporting *catalog*. But, do we need to distinguish the catalog in the first place? My understanding was that we could specify a catalog when making an initial connection to Presto; then, we can just work with schema and table names.

pyongjoo:
Feel free to merge this if you think it's ready.  I'd like to rebase my branch on this.

dongyoungy:
I further fixed presto tests in pyverdict + disabled Impala unit test. I am merging this PR.

PR #380, opened 2019-06-08
Add support for Presto's REGEXP_* functions

### LINKED ISSUE #377: [Presto] Missing REGEX functions
https://prestosql.io/docs/current/functions/regexp.html

I see Verdict has the `REGEXP_CONTAINS` function added but not the Presto equivalent `regexp_like`.

Other missing functions I think:
```
REGEXP_EXTRACT
REGEXP_EXTRACT_ALL
```

I've tried adding them to the following files and compiling:
```
src/main/antlr4/org/verdictdb/parser/VerdictSQLLexer.g4
src/main/antlr4/org/verdictdb/parser/VerdictSQLParser.g4
```
But for some reason it didn't work.

dongyoungy:
(Resolves #377)
Add support for the following REGEXP_* functions in Presto as shown in https://prestosql.io/docs/current/functions/regexp.html:

- REGEXP_EXTRACT_ALL
- REGEXP_EXTRACT
- REGEXP_LIKE
- REGEXP_REPLACE
- REGEXP_SPLIT

Please note that the lambda expression used in one of the example at the website is not currently supported in VerdictDB.


pom.xml
```diff
@@ -300,6 +300,14 @@
             <activation>
                 <jdk>1.7</jdk>
             </activation>
+            <dependencies>
+                <dependency>
+                    <groupId>com.facebook.presto</groupId>
+                    <artifactId>presto-jdbc</artifactId>
+                    <version>0.212</version>
+                    <!-- <scope>test</scope> -->
```
pyongjoo: It looks like we have been including jdbc drivers only for the test phase.  Is there any difference for Presto?

voycey:
That driver version is also very outdated, the project has also forked to
prestosql/presto where the majority of the updates are being done!

dongyoungy:
I think we had an internal discussion of 'providing JDBC drivers' vs. 'letting users use their own JDBC' and IIRC, the test-only scope has been commented out as a result. It was commented out in this commit: https://github.com/mozafari/verdictdb/commit/ac65e2c9a6a9f44a135688b1fc97ad5e47553a73

I will update the JDBC version and restore the test scope for now.

dongyoungy:
After a little big of digging, it looks likes our PrestoJDBCConnection class requires `PrestoStatement` and `QueryStats` so that is actually why we removed 'test' scope previously.

dongyoungy:
I am merging this PR after updating the group id of presto-jdbc to io.prestosql (which should be more up-to-date).

pyongjoo:
Thanks!

commercial-hippie:
Thanks for this!

